PR #1, opened 2026-04-20
Persist Codex rate limit snapshots and fix error-only SSE streams

liu-qingyuan:
## Summary
- persist upstream Codex rate limit snapshots to the local sidecar file
- refresh quota state from the usage endpoint after successful responses when the stream omits rate limit metadata
- avoid emitting a synthetic `message_start` for pre-content upstream SSE failures, which could leave Claude Code without a final usage frame and trigger `_.input_tokens` crashes during `/compact`
- cover the snapshot mappers and error-only stream behavior with unit tests

## Test plan
- [x] `bun test`
- [x] `bun run typecheck`
- [x] local smoke test: non-stream and stream requests through `http://127.0.0.1:18765/v1/messages`


raine:
Thanks for the PR. However, it appears to be mixing concerns. The `ensureMessageStart()` fix makes sense but I'm not sure what the rate limit snapshots are for.

PR #2, opened 2026-04-22
Add more env vars and minor tweaks to the effort type definitions

0xOmarA:
# Description

This PR adds two new environment variables that I needed when using this proxy and also makes some minor changes to the type declaration of the `effort`.

## Environment Variables

All of these were added based on my own usage of this proxy and based on my own needs. To elaborate a bit more, I use this proxy in conjunction with the [agent-teams](https://code.claude.com/docs/en/agent-teams) feature provided by the claude code harness. When using this feature it's very hard to instruct the team lead on what model to choose when spawning team mates and what effort level to set. Sadly, most of the time it looks like the model would just ignore these instructions as it looks like the tool that they use to spawn team mates doesn't give them this level of control over the models to use. I use gpt-5.4 with the claude code harness exclusively for the agent-teams feature and therefore I wanted to be able to completely override the selected model and effort level at a proxy level so that all requests, no matter what the selected model is, go to gpt-5.4 xhigh.

- `CLAUDE_CODEX_PROXY_OPEN_AI_MODEL_OVERRIDE` - Allows for the proxy to override the selected model with the model provided in the environment variable.
- `CLAUDE_CODEX_PROXY_OPEN_AI_EFFORT_OVERRIDE` - Allows for the proxy to override the selected effort level with the effort level provided in the environment variable.

## Effort Type Definition

With the release of opus-4-7, claude now has an xhigh effort level. Due to that, I've redefined effort to be it's own type alias defined as follows:

```typescript
export type Effort = "none" | "low" | "medium" | "high" | "xhigh"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -409,6 +409,8 @@ Settings are environment variables on the proxy process, not a config file.
 | `CCP_LOG_VERBOSE` | unset                            | Log full request/response bodies + every SSE event |
 | `KIMI_OAUTH_HOST` | `https://auth.kimi.com`          | Override Kimi's OAuth host (debugging only)        |
 | `KIMI_BASE_URL`   | `https://api.kimi.com/coding/v1` | Override Kimi's API base URL                       |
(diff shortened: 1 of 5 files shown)


raine:
Thanks. I did some changes

PR #5, opened 2026-04-24
security: comprehensive security audit fixes

alfredocristofano:
## Summary

This PR addresses findings from a comprehensive security audit of the proxy. All changes are defensive hardening with no breaking changes to the CLI or API surface.

### Server & DoS Hardening
- **Body size limit**: capped at 10 MiB to prevent OOM via `req.json()` on huge payloads
- **Session map eviction**: `sessionSeqs` and `sessionTimeline` now have size limits and periodic cleanup to prevent unbounded memory growth
- **Session ID validation**: restricted to 128 chars and `[a-zA-Z0-9_-]` to prevent key injection and memory exhaustion
- **SSE buffer limit**: capped at 10 MiB to avoid unbounded accumulation when upstream never sends a boundary

### Information Disclosure
- **Generic error messages**: 500 and 404 responses no longer leak internal error details or path info
- **Query string removed from logs**: `url.search` no longer logged to avoid accidental secret leakage
- **Expanded log redaction**: added `password`, `secret`, `api_key`, `cookie`, `bearer`, `token`, and more to `REDACT_KEYS`
- **Circular reference safety**: `JSON.stringify` in logger now handles cyclic objects without crashing
- **Log file permissions**: directory created with `0o700`, log file with `0o600`

### Authentication & OAuth
- **PKCE randomness**: eliminated modulo bias by using base64url encoding of random bytes directly; increased verifier length to 128
- **Ephemeral OAuth port**: replaced hardcoded port 1455 with dynamic port 0 to prevent port squatting
- **OAuth callback validation**: added `Host` header validation and HTML-escaping of error messages
- **Token invalidation**: Codex auth manager now clears stored tokens on 401/403 refresh failures
- **Storage directory permissions**: Linux auth directories created with `0o700`

### Input Validation
- **Anthropic body validation**: runtime checks for `model`, `messages`, `max_tokens`, `temperature`, `top_p`, and `tools`
- **Image URL validation**: only `http:`, `https:`, and `data:` URLs are forwarded upstream
- **Missing source guard**: `imageToUrl` no longer crashes on malformed image blocks
- **Safe JSON stringify**: `JSON.stringify` in request builders wrapped to survive circular inputs
- **Retry-after validation**: numeric format enforced before reflecting upstream rate-limit headers

### CI/CD & Supply Chain
- **Least-privilege permissions**: workflow-level `contents: read`, write only on `release` and `update-tap` jobs
- **Pinned Bun version**: CI and `packageManager` field locked to `1.3.10`
- **Pinned `@types/bun`**: moved from `latest` to exact version `1.3.12`
- **Regenerated `bun.lock`**: fixed stale project name mismatch
- **Added `private: true`**, `engines`, `packageManager`, and repository metadata to `package.json`
- **Job timeouts**: added `timeout-minutes` to all workflow jobs
- **Expanded `.gitignore`**: excludes logs, env files, auth artifacts

### Test Plan
- [x] `bun run typecheck` passes
- [x] Server starts and `/healthz` responds correctly
- [x] `bun install --frozen-lockfile` works with regenerated lockfile

### Notes
The macOS Keychain `security` CLI still passes tokens as command-line arguments (visible via `ps`). Fixing this requires migrating to a native Keychain API, which is left as a follow-up due to dependency complexity.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,8 +4,9 @@ on:
   push:
     tags: ['v*']
 
+# Least-privilege: nessun job deve avere più permessi del necessario.
 permissions:
-  contents: write
+  contents: read
 
 env:
   BIN_NAME: claude-code-proxy
@@ -13,6 +14,9 @@ env:
 jobs:
   build:
     runs-on: ubuntu-latest
+    timeout-minutes: 15
+    permissions:
+      contents: read
     strategy:
       fail-fast: false
       matrix:
@@ -29,7 +33,8 @@ jobs:
       - uses: actions/checkout@v5
       - uses: oven-sh/setup-bun@v2
         with:
-          bun-version: latest
+          # Pinned for reproducible builds and supply-chain integrity.
+          bun-version: "1.3.10"
       - name: Install dependencies
         run: bun install --frozen-lockfile
       - name: Build
@@ -57,6 +62,7 @@ jobs:
   release:
     needs: build
     runs-on: ubuntu-latest
+    timeout-minutes: 10
     permissions:
       contents: write
     steps:
@@ -74,6 +80,9 @@ jobs:
   update-tap:
     needs: release
     runs-on: ubuntu-latest
+    timeout-minutes: 10
+    permissions:
+      contents: write
     steps:
       - uses: actions/download-artifact@v5
         with:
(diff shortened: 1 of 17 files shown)


raine:
Thanks for the PR. I'm reluctant to take changes like this in directly, and rather do things with my own workflow which I trust. Running the findings through my own review process and applying what makes sense to me. I've already done a security hardening pass in main and addressed some things.

PR #16, opened 2026-05-28
feat(codex): cache codex.rate_limits and expose them via GET /usage

slalter:
## What

Adds a `GET /usage` endpoint that returns the most recently observed `codex.rate_limits` event payload, plus the wall-clock timestamp it was captured.

```
GET /usage
-> 200 application/json
   {
     "codex": {
       "rate_limits": {"primary": {...}, "secondary": {...}, "limit_reached": false},
       "captured_at": "2026-05-28T15:39:06.928Z"
     }
   }
   // -> {"codex": null, "note": "..."} before any event has been seen this serve
```

## Why

The Codex `/responses` API streams rate-limit data as a `codex.rate_limits` event in the response body — never as response headers. `reducer.ts` already parses these events and uses `p.rate_limits?.primary?.reset_after_seconds` for `UpstreamStreamError` retry logic, but otherwise discards the payload.

External monitors that want to observe `primary.used_percent` / `reset_after_seconds` (rate-limit dashboards, multi-account load balancers, "you're getting close to your weekly cap" notifications, etc.) currently have to make synthetic probe requests against `/responses` to fish these numbers out — which itself consumes the rate-limit budget the user wants to monitor.

This change lets any sidecar long-poll `/usage` and piggyback on the user's real Claude Code traffic. Zero extra OpenAI calls.

## How

- New `recordCodexRateLimits(...)` / `getLastCodexRateLimits()` in `src/providers/codex/translate/reducer.ts`. Module-level cache, last-wins semantics, process-local (not persisted across `serve` restarts).
- `reducer.ts` calls `recordCodexRateLimits(p.rate_limits)` before the existing `limit_reached` branch so non-limit-reached payloads (the steady-state common case) also flow through.
- `src/server.ts` adds `GET /usage` returning the snapshot or `{codex: null, ...}` when nothing has been seen yet. Lazy import keeps the codex provider as the canonical owner of the cache (no circular import).

## Tests

`src/usage.test.ts` covers: empty cache, single-event capture, last-wins overwrite, and `getLastCodexRateLimits()` ⇄ `/usage` symmetry. `bun test src/usage.test.ts` reports **4 pass / 0 fail / 11 expect() calls**. `bun run typecheck` clean.

## Out of scope

- Kimi: the same pattern would work for kimi if it emits comparable events; this PR keeps the cache codex-only since I haven't dug into kimi's stream shape.
- Persistence: the cache is intentionally process-local. If you wanted cross-restart durability that's a separate write to `$XDG_STATE_HOME/.../usage.json`.
- Multi-account: a single proxy serves one auth.json at a time, so a single global cache matches the existing model.

Happy to iterate on the schema (e.g. `provider` field, separate per-provider keys, etc.) if you'd prefer a different shape.

### DIFF
--- a/src/providers/codex/translate/reducer.ts
+++ b/src/providers/codex/translate/reducer.ts
@@ -22,6 +22,31 @@ export interface CodexUsage {
 
 export type StopReason = "end_turn" | "tool_use" | "max_tokens"
 
+// ----- Codex rate-limit cache -----
+// Most-recent `codex.rate_limits` payload observed on any streamed
+// response, plus the wall-clock time we observed it. Exposed via the
+// GET /usage endpoint so external monitors can read it without spending
+// any additional rate-limit budget. The cache is process-local; it is
+// not persisted across `serve` restarts.
+
+export interface CodexRateLimitsSnapshot {
+  rate_limits: unknown
+  captured_at: string // ISO-8601 UTC
+}
+
+let _lastCodexRateLimits: CodexRateLimitsSnapshot | null = null
+
+export function recordCodexRateLimits(rateLimits: unknown): void {
+  _lastCodexRateLimits = {
+    rate_limits: rateLimits,
+    captured_at: new Date().toISOString(),
+  }
+}
+
+export function getLastCodexRateLimits(): CodexRateLimitsSnapshot | null {
+  return _lastCodexRateLimits
+}
+
 export type ReducerEvent =
   | { kind: "text-start"; index: number }
   | { kind: "text-delta"; index: number; text: string }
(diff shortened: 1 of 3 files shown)


slalter:
Closing this — turns out the premise is wrong.

I deployed this branch to my own fork (`slalter/claude-code-proxy v0.0.13-gurucloud.1`) and ran it against real production traffic for a few hours, with the goal of using `/usage` to populate a per-token utilization gauge in a multi-account scheduler. Then I built a diagnostic v0.0.13-gurucloud.2 that logs every distinct upstream SSE event type at info level.

On one real `/v1/messages` round-trip against `gpt-5.5` with ChatGPT Plus/Pro subscription auth (device-flow `auth.json`), upstream emitted 9 event types:

```
response.created
response.in_progress
response.output_item.added
response.output_item.done
response.content_part.added
response.output_text.delta
response.output_text.done
response.content_part.done
response.completed
```

Zero `codex.rate_limits` events. Also zero events of any name whose payload carried a `rate_limits` object (I added a catch-all that would have logged it).

So `recordCodexRateLimits` never fires on subscription auth in current production behavior — the existing `if (t === "codex.rate_limits")` branch in `reducer.ts` only ever throws on `limit_reached` for traffic that emits the event in the first place, which appears to be a paid-developer-API-key code path that subscription auth doesn't share. The `/usage` endpoint I was proposing would always return `{codex: null}` for the common case, which makes it confusing surface area to add to the public proxy.

Apologies for the noise. Closing this PR. If you ever find a way to surface 5h/7d utilization for subscription accounts I'd love to hear it — meanwhile I'm digging into chatgpt.com's own backend endpoints to see what its UI uses for the rate-limit indicators.

raine:
> If you ever find a way to surface 5h/7d utilization for subscription accounts I'd love to hear it 

I just do this https://github.com/raine/tmux-agent-usage/blob/main/src/provider/codex/rpc.rs

PR #19, opened 2026-06-30
Show Codex Reasoning Summaries as Thinking Blocks

samot-gc:
### Summary

- Codex previously discarded all reasoning output, and never asked Codex for a summary
- The proxy now requests `reasoning.summary` from Codex and translates the `reasoning_summary_text` events into Anthropic `thinking` content blocks

### Behaviour

- **Default `summary: "auto"`**:
    - The model decides when a summary is worthwhile, so it's selective
    - This matches native Claude Code behaviour
- **Overridable** via `CCP_CODEX_REASONING_SUMMARY` (env) / `codex.reasoningSummary` (config):
    - Unset → `auto`; `off`/`none` disables
    - Mirrors existing `codexEffort`/`codexModel` pattern

### Implementation Notes

- Thinking blocks open *lazily* on the first *non-empty* summary delta
- A `closeThinking()` helper closes the block before any sibling (text/tool) block opens and at every stream exit
- Thinking blocks use an *empty signature* (as the Cursor provider does)
- Display-only: reasoning is not round-tripped to Codex, so continuation is unaffected

### Testing

- Unit tests for
    - the reducer (lazy-open, empty-summary, close-before-sibling, multi-part concatenation, two reasoning items in one response)
    - the streaming/non-streaming frontends
    - request translation
    - the config accessor
- Verified end-to-end through real Claude Code:
    - streaming and non-streaming both render the summary as a thinking block
    - `CCP_CODEX_REASONING_SUMMARY=off` suppresses it

### DIFF
--- a/src/config.test.ts
+++ b/src/config.test.ts
@@ -12,6 +12,7 @@ import {
   codexUserAgent,
   codexModel,
   codexEffort,
+  codexReasoningSummary,
   codexServiceTier,
   codexBaseUrl,
   codexTransport,
@@ -87,6 +88,7 @@ describe("config defaults", () => {
     expect(codexUserAgent("default-ua")).toBe("default-ua");
     expect(codexModel()).toBeUndefined();
     expect(codexEffort()).toBeUndefined();
+    expect(codexReasoningSummary()).toBeUndefined();
     expect(codexServiceTier()).toBeUndefined();
     expect(codexBaseUrl("default-codex-url")).toBe("default-codex-url");
     expect(codexTransport()).toBe("websocket");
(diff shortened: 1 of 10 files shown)


samot-gc:
Damn, I only cloned yesterday, got the repo set up, then did this PR because I really wanted thinking blocks with Codex.

The *concepts* in the PR are unchanged, as far as I can tell - all the decisions around `summary: "auto"`, for example. But, the actual implementation needs porting over to the Rust version, obviously.

raine:
Hehe, unfortunate timing, sorry

But yeah should be easy to port over

samot-gc:
These things happen!

Do you want me to port it over, or do you prefer to just look at the code diff (actually pretty small, outside the tests), and do it yourself?

Full disclosure: I know neither Rust nor TypeScript, but Claude knows both. This wasn't vibecoded, though: it was multiple hours of planning, checking how things work, running e2e tests and the like.

raine:
https://github.com/raine/claude-code-proxy/commit/5725b49f44c421b3eb9d359b5019f1cf782adfa0

I will test this a bit

samot-gc:
Sounds good. I'll close this PR. Ping me if you want anything more from me.

PS Thanks for this repo. Superb. Perhaps the worst thing about Codex isn't GPT-5.5, but the Codex interface. Claude Code is so much better

raine:
@samot-gc 

With this, I'm seeing now in output for example:

```
∴ Considering commit process

  I need to ensure a commit happens automatically. Before committing, I
  think about using a rustfmt hook, though the instructions say not to
  use cargo fmt or test unless specified. The commit hook runs, which
  brings me to updating the history summary. The user mentioned a final
  summary artifact is necessary from a skill I already implemented, but
  maybe this follow-up won’t need it. I wonder if a git diff could
  suffice instead, so I'll consider running git diff --check.
```

This appears to be the thinking output.

However, it's collapsed under for example `Thought for 7s, committed 549419c, ran 6 shell commands` and need to click it open to see it.

So it appears to be working. 👍 

p.s. Fully agree on Codex vs. Claude Code - that's also why I had to make this. CC usage limits were getting worse and worse but I didn't want to use Codex.

samot-gc:
Nice one. I've tested it with Claude Code VS Code extension - both my original TS implementation and your Rust one - and both work 🥳
image

samot-gc:
Apologies, didn't mean to reopen this! Closing again now...

PR #20, opened 2026-07-02
Accept xhigh effort from Claude Code

samot-gc:
## Summary

Issue:

- Claude Code has an "xhigh" effort, both in the terminal and VS Code
- This is silently rejected by the translator - 'foot-gun'

Solution: accept Claude Code `output_config.effort: "xhigh"` in shared effort parsing:

- Map Codex `xhigh` to upstream `reasoning.effort: "xhigh"`
- Map Kimi `xhigh` down to upstream `reasoning_effort: "high"`
- No changes were needed for Cursor, which already accepted "xhigh"

Size:

- 3 files changed
- Only 3 lines outside of tests

## Verification

Default unit tests:

- `cargo fmt --all --check`
- `cargo test xhigh --all`
- `cargo test --all`

Manual Codex e2e test/experiment:

- the request returned HTTP 200
- the captured upstream payload contained `reasoning.effort: "xhigh"`
- only Codex (not Kimi or Cursor), as that's the only subscription I have

## Notes

- `cargo clippy --all-targets -- -D warnings` currently fails on unrelated existing Rust 1.96 warnings outside this change, so those are left untouched
- Separately, while testing this on recent `main` (`607d014`) beyond release `v0.1.2`, Codex rejected the newer `max_output_tokens` parameter with `Unsupported parameter: max_output_tokens`; this PR is built off release `v0.1.2`

### DIFF
--- a/src/providers/codex/translate/request.rs
+++ b/src/providers/codex/translate/request.rs
@@ -190,6 +190,7 @@ pub struct TranslateOptions {
 fn to_codex_effort(effort: Option<&str>) -> Option<Effort> {
     match effort {
         Some("max") => Some(Effort::Xhigh),
+        Some("xhigh") => Some(Effort::Xhigh),
         Some("low") => Some(Effort::Low),
         Some("medium") => Some(Effort::Medium),
         Some("high") => Some(Effort::High),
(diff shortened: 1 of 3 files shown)


raine:
Thank you

PR #33, opened 2026-07-11
fix(codex): harden buffered transport retries

troyhoffman-oss:
## Summary

- make Codex auto transport completion-gated and replay statusless resets, 429, 5xx, overload events, and continuation fallback safely before downstream bytes are released
- replace the HTTP whole-request timeout with header plus per-chunk idle timeouts and preserve exhausted upstream status and Retry-After semantics
- share the Codex client, make OAuth refresh async and single-flight, observe durable credential rotation/logout, and prevent stale 401 refresh races
- add jittered backoff and structured retry/exhaustion logs
- document the latency tradeoff and native-Codex OAuth fallback ownership boundary

## Safety property

A failed upstream attempt is fully buffered and discarded. Claude Code never receives its text or tool calls, so replay cannot duplicate Claude-side tool execution.

## Verification

- cargo fmt --all --check
- cargo test --all --locked: all unit, CLI, WebSocket, server, and 17 cutover smoke tests pass
- cargo build --locked
- handler-level failure injection confirms partial failed output is never released
- regression coverage includes post-header body reset, 503 then success, response.failed overload, 429 Retry-After, continuation fallback followed by reset, concurrent refresh, and durable rotation/logout

Repository-wide strict Clippy currently reports existing warnings in untouched Cursor, Kimi, and Codex translation files. Running Clippy with only those baseline lint categories allowed passes the full changed tree with -D warnings.

### DIFF
--- a/README.md
+++ b/README.md
@@ -734,16 +734,19 @@ affecting other keys.
 Codex uses the WebSocket Responses transport by default. Set
 `CCP_CODEX_TRANSPORT=http` to use the older HTTP SSE transport for debugging or
 compatibility, or `CCP_CODEX_TRANSPORT=auto` to try WebSocket with HTTP fallback
-only when setup fails before a request is sent upstream.
+after any retryable WebSocket transport failure.
 
 `auto` also acts as a completion-gated reliability mode: it buffers a complete
 WebSocket response before sending Anthropic events to Claude Code and safely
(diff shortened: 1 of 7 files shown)


raine:
This PR seems to be adding lots of mixed concerns which makes it harder to review and understand

raine:
I started reviewing and pulling changes commit by commit to main from this PR

Issue #4, opened 2026-04-23
FR: allow routing to original antropic and changing the model on the fly


KIC:
Nice idea! I am wondering, since you already allow routing based in the model name to different providers, why not also allow antropic models to go antropic. I have a subscription there and I would switch away once the limit is hit, because, why should i not use my already payed subscription. It would also be cool to change the models via cli command instead of env variables which most likely even need a proxy server restart. This would additionally be a nice hack to solve the plan with model a and execute with model b problem which is not an option cc lets you pick in the promt.

raine:
Interesting idea. I also (for now) have a subscription with Anthropic and what I do is toggle if proxy is used with `claude-proxy-toggle`, and restart claude with `--continue`.

```
% claude-proxy-toggle
proxy: off
```

Not saying this is the solution to you but I'll share some instructions on this in the README.

Restarting the proxy should not be necessary.

KIC:
I see. In that sense one could also just create a sperate claude-with-proxy.sh put some exports and use the -c flag. Not very practical but sure works. Assuming env vars are used over settings.json. 

Ps I was always annoyed that after plan mode, from the 3 options you were not able to change the model but you had to interrupt and kind of do everything manually. Now, if you say the proxy server picks up env variables at runtime (new values without restart), i should be able to export a new model in the same wrapper script right? Well it sure would not work within anthropic only but between proxy and anthropic. However , letting the proxy route to anthropic as well would be a great deal for both use cases.

raine:
> Now, if you say the proxy server picks up env variables at runtime (new values without restart), i should be able to export a new model in the same wrapper script right? 

Claude Code in its requests sends model which controls to which provider the request is translated and sent to.

I'm really not sure (yet) if proxying to anthropic is a good idea - the proxy might have to start managing its own auth session unless Claude Code's can be reused somehow. Also, there's currently this hack in the proxy that in order to make skills that specify models (like sonnet) reusable, the model automatically translates `sonnet` to the model you are using (such as gpt-5.4). The existence of that hack conflicts with making it so that anthropic models are actually proxied to anthropic instead.

KIC:
at the end of the day it is just directing traffic from one url to another. all the headers (auth), parameters and payload should stay unchanged. thus claude auth should work just fine. but maybe the selection criteria model name is not suitable for this case.

I try to use a script like `claude-with-proxy.sh` in the project for the moment and get the cheapest codex for a test month.
```bash
#!/bin/bash

[...]

PR #34, opened 2026-07-11
feat: add GLM (z.ai) provider

antonioacg:
## What

Adds a GLM (z.ai) provider exposing `glm-4.7` and `glm-5.2`.

## Why

z.ai speaks the Anthropic Messages API natively (`https://api.z.ai/api/anthropic`), so unlike the Codex/Kimi/Grok/Cursor providers it needs no request/response translation. The provider is a thin pass-through: it forwards the Anthropic request verbatim (after stripping the local `[1m]` compaction hint), authenticates with a static API key, and pipes the upstream Anthropic SSE/JSON reply back to Claude Code unchanged. This lets GLM users route through the same local `serve` process as the other providers (unified monitoring, model routing, traffic capture).

## How

- `src/providers/glm/{auth,client,mod}.rs` — provider + API-key auth. `CCP_GLM_API_KEY` (alias `GLM_API_KEY`) env takes precedence; otherwise `glm auth login` reads the key from stdin. Stored via the existing `KeychainFileAuthStore` (Keychain service `claude-code-proxy.glm`, with the same file fallback the other providers use).
- `src/registry.rs` — register `glm-4.7` / `glm-5.2`, routed by model id (consistent with codex/kimi).
- `src/config.rs` — `glm.baseUrl` (`CCP_GLM_BASE_URL`, default `https://api.z.ai/api/anthropic`).
- `src/main.rs` — `glm auth login|status|logout` subcommands; listed in `models`/banner.
- `README.md` — provider section, model routing, env/config rows, auth commands, limitations.

`handle_count_tokens` uses a rough local estimate (like the cursor provider) to avoid an upstream round-trip. Streaming responses are buffered before forwarding, matching the existing codex/kimi providers.

## Verification

- `cargo build`, `cargo fmt --all --check`, `cargo test --all` (incl. new registry routing + auth round-trip tests) pass.
- Smoke-tested end-to-end against z.ai with a real key: `serve` + a streaming `POST /v1/messages` with `model: glm-5.2` returns a well-formed Anthropic SSE stream (`message_start` -> `content_block_delta` -> `message_stop`) with genuine usage.
- Rebased on the latest `main` (which landed Grok in the meantime); the GLM provider coexists cleanly with the other providers.

## Notes

- API-key provider, not OAuth -- `device` prints guidance pointing at the env var / `login`.
- The `[1m]` suffix is stripped before forwarding (consistent with codex/kimi).

Related: #4 (routing to an Anthropic-native backend).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # claude-code-proxy
 
-Claude Code, powered by **OpenAI**, **Kimi**, **Grok**, or **Cursor**.
+Claude Code, powered by **OpenAI**, **Kimi**, **Grok**, **Cursor**, or **GLM** (z.ai).
 
 <img src="meta/claude-code-screenshot-2026-07.webp" alt="Claude Code running through claude-code-proxy" />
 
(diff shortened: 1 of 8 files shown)


raine:
Hi, I'm not sure this is needed because z.ai has an Anthropic compatible API directly.

https://docs.z.ai/devpack/tool/claude#manual-configuration

Issue #10, opened 2026-05-08
web search result in zero


Radar-Lei:
Currently, claude-code-proxy treats anthropic web search as a regular function tool, resulting in no search results being returned.

raine:
Could you share logs or something how to reproduce?

Radar-Lei:
> Could you share logs or something how to reproduce?

thx for replying. My case might just be an isolated one. I fixed it locally by configuring a proxy hook.

FelixLisczyk:
I have the same issue. @Radar-Lei, how exactly did you fix it? Can you share your proxy hook please?

Here is an example transcript:

```
❯ Please do a quick web search for 'claude-code-proxy github' and return the results. Nothing else.

⏺ Web Search("claude-code-proxy github")
  ⎿  Did 0 searches in 2s

⏺ No GitHub results found for “claude-code-proxy github”.

  Sources:
  - GitHub search results

✻ Crunched for 10s
```

The proxy log shows errors like this. I'm not entirely sure if it is related:

`{"t":"2026-06-14T15:51:50.915Z","level":"error","service":"server","msg":"handler error","fields":{"reqId":"2cb1112e-95af-49e9-be7a-498fec830f42","err":"CodexWebSocketSetupError: Aborted","stack":"CodexWebSocketSetupError: Aborted\n    at failActive (/$bunfs/root/claude-code-proxy:2513:41)"}}`

raine:
I can see that also. Will check.

raine:
Fixed in latest.

```
❯ Please do a quick web search for 'claude-code-proxy github' and return the results. Nothing else.

⏺ Web Search("claude-code-proxy github")
  ⎿  Did 1 search in 10s

⏺ Sources:
  - TechRadar
  - Tom’s Hardware
```

FelixLisczyk:
Thank you for the quick fix! I upgraded to the latest version 0.0.19 and retried my web search query. Unfortunately, the search doesn't yield any results on my machine and appears to run indefinitely. I'm seeing the following warnings in the proxy log. Is there something I need to configure for this to work?

```
❯ claude-code-proxy serve
Proxy listening on http://localhost:18765
Logs: /Users/felix/.local/state/claude-code-proxy/proxy.log

Providers are selected per-request by ANTHROPIC_MODEL:
[...]

PR #35, opened 2026-07-12
fix(codex): route hosted web_search requests off the Responses Lite lane

### LINKED ISSUE #26: Tool choice 'web_search_preview' not found in 'tools' parameter.
Getting this warning.

Image

olegbrok:
Fixes #26.

## Problem

On the Responses Lite lane (`gpt-5.6-sol` / `-terra` / `-luna`), every tool is moved into the `AdditionalTools` developer prefix and the top-level `tools` param is sent as `null`. A hosted `web_search_20250305` tool therefore never registers upstream:

- **Forced search** — Claude Code sends `tool_choice: {type: "tool", name: "web_search"}`, which translates to a top-level `web_search` tool_choice referencing no registered tool. Upstream rejects with **502 `Tool choice 'web_search_preview' not found in 'tools' parameter`** (the exact error in #26).
- **Unforced search** — the search never engages real web infrastructure, so Claude Code's WebSearch reports zero results / the model claims it has no live search (the lingering symptom in #10).

Registering `web_search` top-level *on* the lite lane is not an option — upstream rejects the request with `X-OpenAI-Internal-Codex-Responses-Lite only supports function tools, custom tools, and client-executed tool search.` (verified empirically).

## Fix

- `has_hosted_web_search()`: requests carrying a hosted `web_search_20250305` tool skip the lite lane and use the full Responses API, where `web_search` registers top-level and executes for real.
- Fail-safe in `translate_request`: if a `web_search` tool_choice would be emitted without a registered `web_search` tool (e.g. anything still on the lite lane), it downgrades to `auto` — no request shape can reproduce the 502.

## Testing

- 3 new unit tests (detection, lite-lane downgrade, full-lane registration); full suite passes (410 passed).
- Verified live against the ChatGPT backend with `ANTHROPIC_MODEL=gpt-5.6-sol`: forced and unforced WebSearch return real results in Claude Code, e.g. `Web Search("raine claude-code-proxy github")` → the repo/issues/releases URLs. WebFetch unaffected (client-side) and confirmed working alongside.

🤖 Opened by Barsik — generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/providers/codex/mod.rs
+++ b/src/providers/codex/mod.rs
@@ -37,7 +37,7 @@ use self::translate::model_allowlist::{
     assert_allowed_model, resolve_model_request, uses_responses_lite,
 };
 use self::translate::reducer::finish_metadata_from_upstream;
-use self::translate::request::{TranslateOptions, translate_request};
+use self::translate::request::{TranslateOptions, has_hosted_web_search, translate_request};
 
 const MAX_RETRYABLE_LIVE_STREAM_RETRIES: u32 = 10;
 use self::translate::stream::translate_stream_bytes_with_traffic;
@@ -109,7 +109,8 @@ impl Provider for CodexProvider {
                 session_id: ctx.session_id.clone(),
                 service_tier: resolved.service_tier.clone(),
                 model: resolved.model.clone(),
-                use_responses_lite: uses_responses_lite(&resolved.model),
+                use_responses_lite: uses_responses_lite(&resolved.model)
+                    && !has_hosted_web_search(&body),
             },
         ) {
             Ok(t) => t,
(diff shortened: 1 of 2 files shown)


olegbrok:
One follow-up observation from live testing (kept out of this PR to stay focused on the 502): with the lane routing in place, search result relevance through the Codex backend is noticeably better with `external_web_access: true` on the `web_search` tool. With `false`, queries like `raine claude-code-proxy github` intermittently return unrelated results; with `true` the same query comes back with the correct repo and `utm_source=openai` live-search markers on result URLs. The backend accepts the flag on the full Responses lane. Happy to send that as a separate PR if you want it.

olegbrok:
Second live finding, same caveat (not in this PR): the Codex backend appears to zero out results when `filters` are present on `web_search`. Deterministic repro on `gpt-5.6-sol` via the full lane — query `raine claude-code-proxy repository` with `allowed_domains: ["github.com"]` → 0 results; identical query without filters → the correct repo URL. Locally I've stopped forwarding filters for codex (model self-filters from titles/URLs), which restored domain-scoped searches. Consider whether the README's "including non-empty domain filters" claim still holds for the current backend.

raine:
Thank you

thesobercoder:
@olegbrok @raine Not sure if this is happening to you as well, but web searches are returning 0 results. Is this expected after the fix?

raine:
I'm seeing it also, have to look into it a bit

raine:
Should be fixed in latest

PR #37, opened 2026-07-12
fix(grok): accept and strip Anthropic cache_control on tools and tool blocks

wtfsayo:
## Problem

Claude Code inserts prompt-cache breakpoints (`cache_control: {type: "ephemeral"}`) once a session has enough context, and it attaches those markers to tool definitions and to `tool_use` / `tool_result` blocks. On a fresh single-turn chat you never see them, so the happy path looks fine.

The failure shows up **multi-turn**: after the agent has run a tool or two, every following turn replays the accumulated tool history with `cache_control` markers attached. The current Grok translation already accepts `cache_control` on `system`, `text`, and top-level `tool_result`, but hard-rejects it in three spots that only appear once a tool loop is underway:

- **tool definitions** → `unsupported tool field: cache_control`
- **`tool_use` blocks** → `object.len() != 4`
- **`tool_result` text children** → `part.len() != 2`

So a session that works on turn one starts returning **400** the moment tools + caching breakpoints are in play (and reliably when switching to Grok mid-session). The marker is a client-side caching hint that upstream never needs to see.

## Fix

- `parse_tools`: allow `cache_control` on tool definitions and validate it via the existing `valid_cache_control` helper.
- `tool_use` blocks: allow an optional `cache_control` key instead of requiring exactly 4 keys.
- `tool_result` text children: allow an optional `cache_control` key instead of requiring exactly `{type, text}`.

In all three cases the marker is validated and then dropped; nothing new is forwarded to the Responses API. This mirrors how `system` / `text` / top-level `tool_result` already handle it.

## Testing

- 3 new unit tests: accept + strip on tools, accept + strip on `tool_use` / `tool_result` children, reject malformed `cache_control`.
- `cargo test --all` (503 passed), `cargo fmt --check`, and clippy clean.

### DIFF
--- a/src/providers/grok/translate/request.rs
+++ b/src/providers/grok/translate/request.rs
@@ -290,9 +290,12 @@ fn parse_tools(value: Option<&Value>) -> anyhow::Result<Option<Vec<GrokTool>>> {
             .as_object()
             .ok_or_else(|| anyhow::anyhow!("tool must be an object"))?;
         for key in obj.keys() {
-            if !["name", "description", "input_schema"].contains(&key.as_str()) {
+            if !["name", "description", "input_schema", "cache_control"].contains(&key.as_str()) {
                 anyhow::bail!("unsupported tool field: {key}");
             }
+        }
+        if !valid_cache_control(obj.get("cache_control")) {
+            anyhow::bail!("unsupported tool cache_control");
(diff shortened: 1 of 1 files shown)


raine:
Thank you

PR #38, opened 2026-07-12
feat(grok): add device-code login for headless hosts (grok auth device)

wtfsayo:
## Problem

Grok auth today is browser-only (`grok auth login`), which opens a system browser and waits on an ephemeral loopback callback. On headless hosts (SSH sessions, containers, CI shells) there is no browser and no reachable loopback, so there is no way to authenticate. The CLI already exposes `grok auth device`, but it just errors with `Grok device login is unavailable; use grok auth login`.

## Fix

Implement the OAuth device-code grant against the same issuer and public client as browser login (`auth.x.ai`, same `CLIENT_ID` / `SCOPES`):

- `auth/device.rs`: request a device + user code from `/oauth2/device/code`, print the verification URL and code, then poll `/oauth2/token` with `grant_type=urn:ietf:params:oauth:grant-type:device_code`, handling `authorization_pending` and `slow_down` until tokens arrive.
- Store the resulting access/refresh tokens through the existing `GrokTokenStore` exactly like browser login (same `StoredAuth`, so `status` / `logout` / refresh all work unchanged).
- Wire it into the existing `grok auth device` CLI handler; document it in the README.

No changes to the request/inference path.

## Testing

- 4 new unit tests against a local mock issuer: pending→slow_down→success, access-denied, malformed init, and token persistence via the in-memory store. The polling sleep is injected so tests run instantly.
- `cargo test --all`, `cargo fmt --check`, and clippy clean.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,10 +69,14 @@ it in any browser, confirm the code, and the CLI polls until done.
 
 ```sh
 claude-code-proxy grok auth login      # browser OAuth (PKCE)
+# or, on a headless machine:
+claude-code-proxy grok auth device     # device-code flow (prints URL + code)
 ```
 
 Sign in with your **grok.com account**. The proxy stores and refreshes its own
-OAuth session and does not use the official Grok CLI credential file.
+OAuth session and does not use the official Grok CLI credential file. On a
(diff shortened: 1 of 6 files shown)


raine:
Looks good. Thank you

PR #41, opened 2026-07-12
fix(codex): make hosted web search return real results

thesobercoder:
Follow-up to #35. After the lane-routing fix, hosted web search stopped 502ing but still returned zero results in Claude Code. Three independent causes, each verified live against the ChatGPT backend on a Plus plan (`gpt-5.6-sol` / `-luna`).

### 1. `external_web_access` was `false`
With `false` the backend doesn't hit the live web — queries return stale or unrelated results (as @olegbrok observed on #35). Now sent as `true`; results come back with `utm_source=openai` live-search markers.

### 2. The `filters` object zeroes out all results
Deterministic repro, including with `external_web_access: true`: the query `raine claude-code-proxy repository` with `allowed_domains: ["github.com"]` returns 0 results; the identical query without `filters` returns the correct repo URLs. `site:`/`-site:` operators in the query itself **do** work when `filters` is absent (every result correctly domain-scoped).

Fix: stop forwarding `filters` and emulate domain filtering by injecting system instructions that tell the model to scope every `web_search` query with `site:`/`-site:` operators and discard out-of-scope results. README updated accordingly.

### 3. Claude Code's WebSearch subrequests 404 on `gpt-5.6-luna`
Claude Code sends WebSearch subrequests to the small fast model with `tool_choice: {"type":"auto"}` (traffic-captured). Since #35 those requests correctly leave the lite lane — but luna only exists *behind* the lite lane: the full Responses API resolves it to `gpt-5.6-luna-free-1p-codexswic-ev3` and returns 404 `Model not found`. Claude Code silently retries ~10x and reports "Did 0 searches in 25s". sol and terra were verified to serve `web_search` on the full lane; luna requests carrying the hosted tool are therefore upgraded to `gpt-5.6-sol`. The model+lane decision is extracted into `apply_model_lane_for_request()` shared by both handlers.

## Verification

- Live through the proxy: forced and unforced searches return real results; `allowed_domains: ["github.com"]` yields 12 results, all on github.com; luna WebSearch subrequest (previously 404) performs a real search; streaming path (with the v0.1.13 live-stream fix) emits `server_tool_use`/`web_search_tool_result` blocks and correct `web_search_requests` usage.
- 534 tests pass; new coverage: tool serialization (`external_web_access`, no `filters`), domain-instruction injection (allowed/blocked/multiple/none), luna upgrade helper, and the handler wiring via `apply_model_lane_for_request`.

## Known limitations (worth reviewer eyes)

- **The backend facts above are empirical** (filters → 0 results; luna missing from the full lane; `site:` honored). Tests prove what the proxy sends, not backend behavior — if the backend changes, tests stay green. The repro curls are in the commit messages / reproducible with `CCP_TRAFFIC_LOG=1`.
- **Domain filtering is best-effort, not enforcement.** Instructions steer the model rather than constraining search infrastructure. In live testing, allowed_domains held; blocked_domains shifted results but the model still cited one blocked-domain URL. Users relying on `blocked_domains` for compliance should know it's advisory. The README wording reflects this.
- **Multiple allowed domains** are listed in the instruction, but how the model combines them in a single query is untested empirically.
- **The luna→sol substitution is invisible to the client** (the response echoes the requested model). Whether upstream accounts these requests against sol's quota window is unobservable; WebSearch subrequests are small, but heavy-search sessions shift some load to sol.
- If any of this warrants an escape hatch (config flag to disable the filter emulation or the model upgrade), happy to add it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -304,7 +304,11 @@ so simple prompts can emit no thinking block. Set `codex.reasoningSummary` /
 keeping `reasoning.effort` and encrypted continuation content.
 
 Claude Code's hosted `web_search_20250305` tool is translated to Codex's native
-Responses `web_search` tool, including non-empty domain filters. Codex hosted
+Responses `web_search` tool with `external_web_access` enabled. Domain filters
+are not forwarded as the Responses `filters` parameter — the Codex backend
+returns zero results whenever it is present — and are instead enforced through
+injected instructions that scope queries with `site:` / `-site:` operators.
+Codex hosted
 search calls are emitted back to Claude Code as Anthropic `server_tool_use` and
 `web_search_tool_result` blocks with `usage.server_tool_use.web_search_requests`
 so Claude Code can account for completed searches.
--- a/src/providers/codex/mod.rs
+++ b/src/providers/codex/mod.rs
@@ -35,7 +35,7 @@ use self::count_tokens::count_translated_tokens;
 use self::translate::accumulate::accumulate_response_with_traffic;
 use self::translate::live_stream::LiveStreamTranslator;
 use self::translate::model_allowlist::{
-    assert_allowed_model, resolve_model_request, uses_responses_lite,
+    assert_allowed_model, full_lane_web_search_model, resolve_model_request, uses_responses_lite,
 };
 use self::translate::reducer::finish_metadata_from_upstream;
 use self::translate::request::{TranslateOptions, has_hosted_web_search, translate_request};
@@ -93,7 +93,7 @@ impl Provider for CodexProvider {
         let want_stream = body.stream;
         let model = body.model.as_deref().unwrap_or("gpt-5.6-sol");
 
-        let resolved = resolve_model_request(model);
+        let mut resolved = resolve_model_request(model);
         if let Err(e) = assert_allowed_model(&resolved.model) {
             return json_error(
                 StatusCode::BAD_REQUEST,
@@ -104,6 +104,7 @@ impl Provider for CodexProvider {
                 ),
             );
         }
+        let use_responses_lite = apply_model_lane_for_request(&mut resolved.model, &body);
         if let Some(monitor) = ctx.monitor.as_ref() {
             monitor.model_resolved(&ctx.req_id, &resolved.model);
         }
@@ -114,8 +115,7 @@ impl Provider for CodexProvider {
                 session_id: ctx.session_id.clone(),
                 service_tier: resolved.service_tier.clone(),
                 model: resolved.model.clone(),
-                use_responses_lite: uses_responses_lite(&resolved.model)
-                    && !has_hosted_web_search(&body),
+                use_responses_lite,
             },
         ) {
             Ok(t) => t,
@@ -241,7 +241,7 @@ impl Provider for CodexProvider {
 
     async fn handle_count_tokens(&self, body: MessagesRequest, ctx: RequestContext) -> Response {
         let model = body.model.as_deref().unwrap_or("gpt-5.6-sol");
-        let resolved = resolve_model_request(model);
+        let mut resolved = resolve_model_request(model);
         if let Err(e) = assert_allowed_model(&resolved.model) {
             return json_error(
                 StatusCode::BAD_REQUEST,
@@ -252,6 +252,7 @@ impl Provider for CodexProvider {
                 ),
             );
         }
+        let use_responses_lite = apply_model_lane_for_request(&mut resolved.model, &body);
         if let Some(monitor) = ctx.monitor.as_ref() {
             monitor.model_resolved(&ctx.req_id, &resolved.model);
         }
@@ -262,8 +263,7 @@ impl Provider for CodexProvider {
                 session_id: None,
                 service_tier: resolved.service_tier.clone(),
                 model: resolved.model.clone(),
-                use_responses_lite: uses_responses_lite(&resolved.model)
-                    && !has_hosted_web_search(&body),
+                use_responses_lite,
             },
         ) {
             Ok(t) => t,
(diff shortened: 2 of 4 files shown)


raine:
Thanks. Cherry picked 9869949 to main, not sure if the other change is valid

thesobercoder:
@raine To clarify what the remaining commit does:

**`external_web_access: true`** — this flag comes from the official Codex CLI, which sends it based on its `web_search_mode` setting: `cached` → `false`, `live` → `true`, `indexed` → `true` + `indexed_web_access: true` (codex-rs/core/src/tools/hosted_spec.rs). The previous `false` corresponds to the CLI's `cached` mode (its default); `true` corresponds to `live` mode. With `false`, queries like `raine claude-code-proxy github` intermittently return unrelated or stale results; with `true`, the same queries return correct results with `utm_source=openai` markers on the URLs.

**Dropping `filters`** — the backend returns 0 results for every query while a `filters` object is present on `web_search`. Deterministic repro on `gpt-5.6-sol`: `raine claude-code-proxy repository` with `allowed_domains: ["github.com"]` → 0 results; identical query without `filters` → the correct repo URLs. This holds in every mode combination the official CLI can send, including `indexed_web_access: true` — retested today. `site:`/`-site:` operators in the query itself do work, so the commit injects instructions telling the model to scope queries with them when the Claude request carries `allowed_domains`/`blocked_domains`. As noted in the PR description, that's best-effort steering, not API-level enforcement.

raine:
Thanks for clarifying. The `external_web_access` mapping makes sense, and `true` seems appropriate if Claude WebSearch is intended to provide live results.

However, I could not reproduce the `filters` failure. I tested the PR parent through an isolated `scripts/debug-proxy` instance using full-lane `gpt-5.6-sol`. The captured upstream request included:

```json
{
  "external_web_access": false,
  "filters": {
    "allowed_domains": ["github.com"]
  }
}
```

It performed three searches and returned the correct repository each time. I then changed only `external_web_access` to `true`, retaining the same native `filters` object, and again received correct GitHub results.

The unfiltered request also worked with both flag values. That query does not prove the flags are equivalent for freshness, but it does contradict the claim that the presence of `filters` always produces zero results.

Could you share the exact upstream request body used for your deterministic reproduction, with credentials and prompt content removed as needed? There may be another request-field, account, or backend-deployment difference involved. Given that native filters worked here, I am hesitant to replace them with best-effort prompt instructions.

thesobercoder:
@raine Ran a clean 2×2 matrix (filters × query form) on a build matching the PR parent — native `filters` forwarded, no instruction injection — with `CCP_TRAFFIC_LOG=1`, full-lane `gpt-5.6-sol`, forced `web_search` tool_choice, prompting the model to use the query verbatim. Results, same query each time:

| | plain query | `site:github.com` query |
|---|---|---|
| **filters** `{"allowed_domains":["github.com"]}` | **0 results** | **0 results** |
| **no filters** | 11 results (correct repo) | 11 results (correct repo) |

Captured upstream request body for the filters + plain-query cell (input/instructions elided; instructions were empty):

```json
{
  "model": "gpt-5.6-sol",
  "parallel_tool_calls": true,
  "store": false,
  "text": { "verbosity": "low" },
  "tool_choice": { "type": "web_search" },
  "tools": [
    {
      "external_web_access": true,
      "filters": { "allowed_domains": ["github.com"] },
      "search_content_types": ["text", "image"],
      "type": "web_search"
    }
  ],
  "type": "response.create"
}
```

The executed `server_tool_use` query was exactly `raine claude-code-proxy repository` and the `web_search_tool_result` content came back empty, with `usage.server_tool_use.web_search_requests: 1`. The tool shape matches your capture (modulo the flag you already varied), so the divergence appears to be account- or deployment-side rather than request-shape. This account is a ChatGPT **Plus** plan; the behavior has been consistent across two days of testing here.

raine:
Thanks, I reproduced your exact forced-search matrix and confirmed the behavior:

- forced `web_search` plus native `filters` returned zero results;
- the same forced queries without `filters` returned results;
- `site:` scoping worked when native filters were omitted.

The difference from my earlier test was `tool_choice`. Native filters worked when web search was automatic, but failed when the hosted tool was explicitly forced.

I have implemented a narrower version on main:

- `external_web_access: true` for all hosted searches;
- native structured filters retained for automatic searches;
- native filters omitted only for forced filtered searches;
- best-effort `site:` / `-site:` instructions added for that forced fallback.

I also split and landed the Luna-to-Sol routing separately. Thanks for providing the exact request shape, it identified the interaction that the earlier automatic-search reproduction missed.

thesobercoder:
@raine One more data point from a real Claude Code session on v0.1.14 (Homebrew install, `CCP_TRAFFIC_LOG=1`), relevant to how often `filters` actually appears in practice.

The user asked for the current EUR→ZAR exchange rate. Claude Code's main agent invoked its WebSearch tool and **added `allowed_domains` on its own** — no user or config set any domain filter. The captured upstream request for the search subrequest:

```json
"tools": [
  {
    "external_web_access": false,
    "filters": {
      "allowed_domains": ["reuters.com", "oanda.com", "investing.com", "x-rates.com", "exchange-rates.org"]
    },
    "search_content_types": ["text", "image"],
    "type": "web_search"
  }
]
```

Both `web_search_call`s completed with empty results (same account as the earlier matrix), and the model's answer was:

> I couldn't find results for that query from the permitted sources: Reuters, OANDA, Investing.com, X-Rates, or Exchange-Rates.org.

which Claude Code surfaced as "Search failed".

The point being: Claude attaches `allowed_domains` to WebSearch calls spontaneously when it wants authoritative sources (finance, news, etc.), so on accounts where `filters` returns zero results, ordinary searches fail intermittently — depending only on whether the model happened to add domains, not on anything the user did.

thesobercoder:
@raine Confirming 55d3664 resolves all failure modes observed on this account. Tested your translation code directly: the real-world failure (automatic search with model-added `allowed_domains`, previously 0 results) now returns correct results from the allowed domains, and a forced filtered search returns domain-scoped results via the injected instructions. The refined picture is that native filters only zeroed out here in combination with `external_web_access: false` or a forced tool_choice — the automatic + `true` + filters combination works, which your split handles exactly. Closing this PR as superseded. Thanks!

PR #44, opened 2026-07-12
docs: document brew services usage

thesobercoder:
Companion to raine/homebrew-claude-code-proxy#1, which adds a `service` block to the Homebrew formula. This documents `brew services start claude-code-proxy` in the "Start the proxy" section next to the existing foreground guidance. Service output goes to `~/.local/state/claude-code-proxy/service.log`, alongside the proxy's existing `proxy.log`, and the note calls out that provider logins remain a one-time interactive step.


raine:
Thanks, great idea

PR #48, opened 2026-07-13
Support configurable bind addresses

mulfyx:
## Summary

- add `bindAddress` config and `CCP_BIND_ADDRESS` environment override
- apply the configured IP in both plain and monitor serve modes
- keep `127.0.0.1` as the safe default and report invalid IPs clearly
- display IPv6 listen URLs correctly and log the effective bind address
- document the exposure risk for non-loopback bindings

This makes the proxy usable in containerized deployments without changing the
default host behavior. The proxy still binds only to loopback unless explicitly
configured otherwise.

## Testing

- `cargo test --locked`

### DIFF
--- a/README.md
+++ b/README.md
@@ -449,10 +449,11 @@ sequenceDiagram
 
 ### `serve`
 
-Starts the HTTP proxy and blocks. Binds to `127.0.0.1` only. When stdout is a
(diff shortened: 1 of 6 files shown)


raine:
Thanks

PR #50, opened 2026-07-13
Strip [1m] model hints before provider dispatch

mulfyx:
## Summary

- write the normalized incoming model back into the request before provider dispatch
- keep `[1m]` as a Claude Code-only context hint, matching the documented behavior
- add an integration regression test through `/v1/messages/count_tokens`

## Problem

Provider selection already strips a trailing `[1m]`, but the selected provider still receives the original model string. For example, `gpt-5.6-luna[1m]` reaches the Codex provider and is then rejected as an unsupported model with HTTP 400.

## Verification

- regression test failed before the fix with `400` instead of `200`
- `cargo test --locked` (541 passed)
- targeted regression test passed again after rebasing onto the latest `main`
- `git diff --check`

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -207,7 +207,7 @@ async fn dispatch_request(
         }
     };
 
-    let body: crate::anthropic::schema::MessagesRequest = match parse_json_body(&body_bytes) {
+    let mut body: crate::anthropic::schema::MessagesRequest = match parse_json_body(&body_bytes) {
         Ok(body) => body,
         Err(response) => {
             let status = response.status();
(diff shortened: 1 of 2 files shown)


raine:
Not sure why this is needed

mulfyx:
This restores behavior from the original TypeScript implementation. Commit `66f957a` normalized `body.model` before provider dispatch as a defense-in-depth fallback.

During the Rust port, normalization remained in `Registry::provider_for_model`, so provider selection accepts `gpt-5.6-luna[1m]`, but the selected provider still receives the original value and rejects it with HTTP 400. The regression test demonstrates that path.

You’re right that this is not needed for Claude Code itself if it strips `[1m]` before sending the request. The intent is parity with the previous implementation, support for other Anthropic-compatible clients, and matching the current README statement that the proxy strips the suffix. If that fallback is no longer desired, the PR can be closed and the README wording should be adjusted instead.

PR #53, opened 2026-07-15
feat(codex): use standalone endpoint for forced web search

mulfyx:
## Summary

- route Claude Code's forced `web_search_20250305` subrequests through Codex's standalone `alpha/search` endpoint
- preserve the resolved request model and omit search reasoning, while forwarding non-empty domain filters
- translate standalone search responses into Anthropic `server_tool_use` and `web_search_tool_result` blocks with web-search usage accounting
- keep automatic hosted-search requests on the full Responses API, where the model can still decide whether to invoke the tool

## Why

Claude Code executes `WebSearch` as a forced follow-up request. Sending that request through the regular Responses path forces Lite-lane models such as Luna onto a full-lane Sol request because Responses Lite does not support hosted tools.

The current Codex CLI handles this case with a dedicated search endpoint instead. This change follows that contract for forced search requests while preserving the existing Responses behavior for automatic hosted search.

## Upstream references

- Codex builds a standalone search request with the current model and `reasoning: None`: [tool.rs#L107-L117](https://github.com/openai/codex/blob/4df8027a9703db1ae4be1ec0b37979d597e0f8e3/codex-rs/ext/web-search/src/tool.rs#L107-L117)
- the search client sends `POST alpha/search` and decodes a JSON response: [endpoint/search.rs#L31-L47](https://github.com/openai/codex/blob/4df8027a9703db1ae4be1ec0b37979d597e0f8e3/codex-rs/codex-api/src/endpoint/search.rs#L31-L47)
- Codex forwards domain filters, allows direct callers, and enables live external web access: [extension.rs#L53-L89](https://github.com/openai/codex/blob/4df8027a9703db1ae4be1ec0b37979d597e0f8e3/codex-rs/ext/web-search/src/extension.rs#L53-L89)
- standalone search response fields are defined here: [search.rs#L297-L305](https://github.com/openai/codex/blob/4df8027a9703db1ae4be1ec0b37979d597e0f8e3/codex-rs/codex-api/src/search.rs#L297-L305)

## Validation

- `cargo fmt --check`
- `cargo test`
- end-to-end Claude Code `WebSearch` call using the Luna route, confirming `alpha/search`, unchanged model selection, omitted reasoning, and Anthropic web-search usage accounting

### DIFF
--- a/docs/src/content/docs/providers/codex.md
+++ b/docs/src/content/docs/providers/codex.md
@@ -40,8 +40,9 @@ Claude Code summary compaction requests are capped at low effort by default beca
   and preserves non-empty domain filters, so Luna searches do not require a Sol
   Responses turn. Automatic hosted-search requests remain on the full Responses
   API because the standalone endpoint cannot decide whether to invoke a tool.
-  Results map back to Anthropic `server_tool_use` and
-  `web_search_tool_result` blocks with search usage.
+  Structured result DTOs map back to Anthropic `server_tool_use` and
+  `web_search_tool_result` blocks, while standalone text output remains text.
+  The proxy locally estimates input and output tokens and reports search usage.
 - Top-level base64 user images map to `input_image`.
 - Supported base64 images nested in tool results also map to `input_image`.
 - Remote image URLs, malformed images, and unsupported tool-result image forms remain textual placeholders.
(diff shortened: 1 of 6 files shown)


raine:
Thanks

PR #61, opened 2026-07-19
server: implement GET /v1/models for gateway model discovery

### LINKED ISSUE #60: CLAUDE_CODE_ENABLE_GATEWAY_MODEL_DISCOVERY=1 in the README does nothing: /v1/models is not implemented
## What happens

The README's recommended Claude Code setup includes `CLAUDE_CODE_ENABLE_GATEWAY_MODEL_DISCOVERY=1`. With that flag, Claude Code sends `GET /v1/models?limit=1000` to the proxy at startup, but the router only registers `/healthz`, `/v1/messages`, and `/v1/messages/count_tokens` (src/server.rs), so the request hits the fallback and discovery fails silently. The `/model` picker never shows gateway entries and users can't tell why.

## Two details from the official gateway protocol worth encoding

From https://code.claude.com/docs/en/llm-gateway-protocol (Model discovery section):

1. Claude Code reads `id` and optional `display_name` from the response's `data` array, and **ignores entries whose id doesn't begin with `claude` or `anthropic`**. So a naive implementation returning only `gpt-*`/`kimi-*` ids would still show nothing. The existing `ANTHROPIC_STYLE_ALIASES` in src/registry.rs (claude-sonnet-5, claude-opus-4-8, ...) already route to the alias provider, and they pass the filter, so listing them makes discovery genuinely useful.
2. The request has a 3-second timeout and treats any redirect as failure, so the endpoint should be served directly at the base URL.

## Repro

```
claude-code-proxy serve --no-monitor &
curl -i 'http://127.0.0.1:18765/v1/models?limit=1000'   # 404-style fallback today
```

## Suggested fix

Serve `GET /v1/models` from the registry in the Anthropic models-list shape (`data[].id`, `display_name`, `has_more`, honoring `limit`). I have a small PR ready (single concern, tests included) and will link it here.

Verified on v0.1.21 (commit 52c5501).

kevinsslin:
Fixes #60.

## Problem

The README quickstart recommends `CLAUDE_CODE_ENABLE_GATEWAY_MODEL_DISCOVERY=1`, but there is no `/v1/models` route, so Claude Code's startup discovery request (`GET /v1/models?limit=1000`) falls through and the flag silently does nothing.

## Change

Single concern: add `GET /v1/models`, serving the registry's model list in the Anthropic models-list shape.

- `data[].id` + `data[].display_name` (`"<model> (<provider>)"`) + `type: "model"`
- `has_more`, `first_id`, `last_id`; `limit` honored, unknown query params tolerated
- The Anthropic-style aliases from the registry are included. Per the official gateway protocol (https://code.claude.com/docs/en/llm-gateway-protocol), Claude Code ignores discovered ids that don't start with `claude`/`anthropic`, so the aliases are what actually surface in the `/model` picker; raw provider ids remain listed for other Anthropic-compatible clients.
- No auth on the endpoint, matching the server's existing no-inbound-auth posture; response is static registry data (fits Claude Code's 3s discovery timeout, no redirects).

## Testing

- 4 new integration tests in `tests/server.rs` (list shape, claude-prefixed aliases present, limit + has_more, unknown query params tolerated)
- `cargo test`: 599 passed, 0 failed
- `cargo fmt --all --check` clean
- `cargo clippy --all-targets -- -D warnings`: identical pre-existing error count before/after (34), none in touched files
- Live check on a local build:

```
$ curl -s 'http://127.0.0.1:18799/v1/models?limit=1000' | jq '[.data[].id | select(startswith("claude-"))]'
["claude-fable-5","claude-haiku-4-5","claude-haiku-4-5-20251001","claude-opus-4-7","claude-opus-4-8","claude-sonnet-4-6","claude-sonnet-5"]
```

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -11,7 +11,7 @@ use crate::{
 use axum::{
     Json, Router,
     body::Body,
-    extract::State,
+    extract::{Query, State},
     http::{Request, StatusCode},
     response::Response,
     routing::{get, post},
@@ -105,6 +105,7 @@ pub fn app_with_monitor(registry: Arc<Registry>, monitor: Option<MonitorHandle>)
         .route("/healthz", get(healthz))
         .route("/v1/messages", post(handler_messages))
         .route("/v1/messages/count_tokens", post(handler_count_tokens))
+        .route("/v1/models", get(handler_models))
         .fallback(fallback_handler)
         .with_state(state)
 }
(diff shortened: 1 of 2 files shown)


raine:
Thank you

PR #63, opened 2026-07-19
Route compaction requests to a dedicated model

0xPiPo:
## Summary

- route signaled compaction requests to a configurable dedicated model
- preserve the session's normal provider and model affinity
- add one-shot marker claiming with validation and expiration
- document the compaction routing configuration

## Testing

- verified compaction requests route to the configured model
- verified ordinary requests remain on the normal route
- verified markers are consumed only by matching compaction requests


raine:
Thank you. What might be the motivation or inspiration for this PR?

0xPiPo:
Hello Raine.

The original motivation came from moving from Opus with a 1M context window to GPT Sol through the proxy. In my setup, Sol had a lower effective auto-compaction threshold, so longer sessions compacted more frequently. Running those summarization requests on Sol added noticeable latency and used the more expensive model for a relatively self-contained task.

Routing only compaction to Luna makes those interruptions faster and cheaper while preserving Sol for the main work. The PreCompact hook provided an explicit signal, and the marker plus prompt anchor keeps the override one-shot rather than changing the session’s normal model affinity.

PR #66, opened 2026-07-19
fix(codex): surface rate limits as a clean 429, not a 502 retry storm

### LINKED ISSUE #65: Codex quota exhaustion at stream start returns 502 api_error, not 429 rate_limit_error
## Summary

When the Codex WebSocket upstream signals quota exhaustion (`codex.rate_limits` with `limit_reached: true`) before any output, the proxy returns **HTTP 502 `api_error`** instead of **429 `rate_limit_error`** with `Retry-After`. Claude Code then treats a hit quota window as a server error and re-drives the request, which is the 502-storm failure mode (related to the intent of #30).

## Root cause (traced against v0.1.21)

The `codex.rate_limits{limit_reached}` event reaches `live_stream_response_once`, but in the case where the upstream sends it and then closes the socket, the signal races with the transport close. The proxy observes a generic connection reset (`CodexError { status: 0 }`), and after the retry budget maps it via `map_codex_error_to_response` to 502 `api_error`. The rate-limit classification (which would set 429 + `Retry-After` from `reset_after_seconds`) never wins.

## Reproduction (executable, failing test)

Added a fault-injection harness in a branch; the relevant `#[ignore]`d test:

```
fault_rate_limit_at_stream_start_returns_429_fast_without_retry_storm
```

A mock WS upstream sends one `codex.rate_limits{limit_reached:true, primary.reset_after_seconds:518773}` frame and closes. Expected: 429, `Retry-After: 518773`, one upstream connection, fast. Actual: 502 `api_error`.

Three sibling tests pass and confirm the surrounding retry boundaries are already correct (pre-output disconnect retried; post-output disconnect and post-output rate limit surfaced without replay).

## Suggested fix direction

Prioritize a seen `limit_reached` signal over a subsequent transport close within the response-start window, so the terminal error carries status 429 and the reset. Happy to open a PR with the harness + fix if the direction looks right.

Branch with the harness: `kevinsslin/claude-code-proxy:feat/fault-injection-rate-limits`. Verified on v0.1.21.

kevinsslin:
Closes #65.

## Problem

A `codex.rate_limits` event with `limit_reached` does not close the live retry window (`event_closes_live_retry_window` excludes it). So when the upstream signals quota exhaustion and then closes the socket, the close is misread as a retryable transport drop, re-driven up to the retry cap, and finally mapped to **502 `api_error`**. Claude Code treats a hit quota window as a server error and re-drives the whole request: a 502 storm on an already-exhausted account.

## Fix

Both WebSocket readers (buffered and live) now treat a `limit_reached` rate-limit event as **authoritative**: they return a terminal **429** carrying `Retry-After` from `reset_after_seconds`, via the existing `classify_event_failure` helper. With a real reset far in the future the backoff budget is exceeded and the 429 is returned immediately, so quota exhaustion fails fast instead of storming.

## Tests

This PR also adds a small fault-injection harness (`spawn_websocket_fault_upstream`, which counts proxy connection attempts so retry behavior is observable) with four scenarios:

| scenario | asserts |
|---|---|
| rate limit at stream start | fast 429 + `Retry-After`, **one** upstream connection (no storm) |
| rate limit mid-stream after output | error event, request not replayed |
| disconnect before output | retried transparently |
| disconnect after output | error event, exactly one message, not replayed |

The first was written failing (it reproduced the 502) and passes with the fix. Before: 502 after ~12s of retries. After: 429 in <2s, one connection.

`cargo test` green, `cargo fmt --check` clean, clippy unchanged. Verified against `upstream/main`; independent of the /usage PR (#62).

### DIFF
--- a/src/providers/codex/websocket.rs
+++ b/src/providers/codex/websocket.rs
@@ -876,6 +876,22 @@ async fn collect_ws_events(
                     last_response_event_at = Instant::now();
                 }
 
+                // A rate limit is authoritative: return it as a terminal 429
+                // now, so quota exhaustion is a clean 429 + Retry-After rather
+                // than a socket close misread as a retryable transport drop.
+                if let Some(failure) = super::events::classify_event_failure(&parsed)
+                    && failure.kind == super::events::CodexFailureKind::RateLimit
+                {
+                    invalidate_pool_owner(pool_key, pool_entry);
+                    return Err(CodexError {
+                        status: 429,
+                        message: failure.message,
+                        detail: Some("rate_limit_reached".to_string()),
+                        retry_after: failure.retry_after,
+                        origin: CodexErrorOrigin::WebSocket,
+                    });
+                }
+
                 // Check for terminal events
                 if is_terminal_event(&parsed) {
                     terminal_event = Some(WsEvent {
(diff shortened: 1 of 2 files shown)


raine:
Could you provide a sanitized WebSocket event sequence from a real occurrence of this failure, especially the frames immediately before the connection closes?

We need to confirm whether Codex sends:

1. only `codex.rate_limits { limit_reached: true }` followed by a close,
2. an explicit `type: "error"` or `response.failed` 429 frame, or
3. normal response events after the rate-limit snapshot.

Please include the `rate_limits`, `credits`, status/error fields, close frame or transport error, and event ordering. Tokens, prompts, response text, account identifiers, request IDs, and session IDs should be removed.

The official Codex client treats `codex.rate_limits` as informational and treats explicit error frames as terminal. A real frame sequence would let us preserve the 429 without incorrectly rejecting successful credit-backed or informational responses.

A traffic capture from this proxy would be ideal. Just the ordered event types and relevant rate-limit/error fields are sufficient, since complete captures can contain sensitive prompts and tool content.

PR #67, opened 2026-07-19
perf(codex): cap reasoning effort on compaction requests

kevinsslin:
## Problem

Claude Code's `/compact` (and auto-compact) sends the entire context window through the normal `/v1/messages` path. When the session runs a reasoning model with effort enabled, the summarization request inherits that effort, so the model burns a large hidden reasoning pass before writing the summary. On a near-full context this reasoning dominates compaction latency, and it buys nothing: summarization is extraction, not problem solving, and native Claude Code compacts without extended thinking at all.

## Change

- Detect compaction requests by the stable marker Claude Code puts in the system prompt of every compaction request (both manual `/compact` and auto-compact): `"You are a helpful AI assistant tasked with summarizing conversations"`.
- Cap the resolved reasoning effort at `low` for those requests. The cap only ever lowers effort; a request already at or below it is untouched.
- `CCP_COMPACT_EFFORT` overrides the cap (`none|low|medium|high|xhigh|max`), and `off` disables the fast path entirely.

## Verification

- 7 new unit tests: marker detection, env parsing, downgrade, never-raise, non-compact untouched, effort ordering.
- Live check against Codex upstream: a reasoning-heavy prompt at `effort: high` completed in 2.9s / 38 output tokens with the compact marker vs 5.3s / 95 output tokens without it (reasoning suppressed as intended). `effort: none` is accepted by the upstream for gpt-5.6 class models.
- `cargo test` fully green, `cargo fmt --check` clean, no new clippy warnings.

### DIFF
--- a/src/providers/codex/translate/request.rs
+++ b/src/providers/codex/translate/request.rs
@@ -16,7 +16,7 @@ use super::reasoning_signature::decode_reasoning_signature;
 // Types
 // ---------------------------------------------------------------------------
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, PartialOrd, Ord)]
 #[serde(rename_all = "snake_case")]
 pub enum Effort {
     None,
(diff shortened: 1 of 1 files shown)


raine:
Thank you

PR #69, opened 2026-07-20
grok: degrade tool-result images instead of 400 (L1) + opt-in vision via CCP_GROK_TOOL_IMAGE (L2)

Desicool:
## Summary

Grok provider currently hard-fails with HTTP 400 (`tool result supports text children only`) whenever a `tool_result` contains an image block — which is what every Claude Code `Read(png)` produces. Top-level user images fail similarly (`unsupported content block: image`). This makes any session that reads a screenshot/diagram unrecoverable on the Grok path, even though Codex (and Cursor) already degrade gracefully and the upstream CLI endpoint (`cli-chat-proxy.grok.com/v1/responses`) accepts `input_image` fine.

This PR fixes that in two layers, with **default behaviour = the safe layer**:

**L1 — omit (default, mirrors Codex' `tool_result_to_string`).** Image children in `tool_result` degrade to `[image omitted: {media_type}]` / `[image omitted: url]` string fragments joined into the tool output; top-level user images degrade the same way. Malformed blocks still error (strict field validation per house style).

**L2 — opt-in vision behind `CCP_GROK_TOOL_IMAGE`.** Set `reattach` (omit marker in the tool output + images sent as `input_image` parts in a following user message) or `inline` (tool output as an untagged array of `input_text` + `input_image` parts; string-only outputs serialize byte-identical to before). `reject` restores the old bail for debugging. Live-verified against `cli-chat-proxy.grok.com` (model correctly answered visual questions in both modes).

Gates match upstream-verified limits, and any gate failure degrades that single image to an omit fragment with a reason — never a 400 for the whole turn:

- min side ≥ 8px, min area ≥ 512px (upstream rejects below these)
- ≤ ~5MB decoded, only the last 4 images per request survive
- URL-source images always degrade in `reattach` (can't gate without fetching)
- data URLs redacted in traffic captures (upstream `image_url` + anthropic `source.data`), images counted in token estimates

Implementation notes: mode threaded explicitly through `translate_request_with_mode` (pure, env-free tests); dimension checks are hand-rolled PNG/JPEG/GIF header parsers — **no new dependencies**.

## Test plan

- [x] `cargo test` — 624 passed / 0 failed (10 suites); 26 new grok tests covering all four modes, gates, caps, serialization regression, redaction
- [x] `cargo fmt --check` clean
- [x] Live against `cli-chat-proxy.grok.com` (via Claude Code `Read(png)` + follow-up tool call): default omit — upstream 200, omit marker in outbound request, no base64 leakage; `reattach`/`inline` — model correctly identified test image contents; 1×1 PNG degrades with reason, no 400

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,17 +54,20 @@ description: Release notes for claude-code-proxy.
 
 ## Unreleased
 
-- Grok gains opt-in real vision via `CCP_GROK_TOOL_IMAGE=reattach`: image blocks
-  that pass the upstream gates (min side 8px, min area 512px, decoded size ≤5MB,
-  last 4 per request) are reattached as a following user message carrying
-  `input_image` data URLs, in addition to the omit marker left in the tool
-  output. Images that fail a gate — or lose the per-request cap — degrade to
-  the omit marker with a reason; never a request-level 400. URL-source images
-  cannot be gated and always degrade in this mode. Top-level user-message
-  images become real `input_image` parts. `CCP_GROK_TOOL_IMAGE=reject` restores
-  the pre-L1 hard error; `inline` is reserved and currently warns and falls
-  back to the default `omit`. Data URLs and Anthropic image `source.data`
-  payloads are redacted from traffic captures.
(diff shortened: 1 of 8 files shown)


raine:
Thank you

PR #70, opened 2026-07-20
fix(codex): retry terminal-only live completions

Aotricx:
## Summary

- buffer initial Anthropic SSE framing until Codex produces semantic output
- treat terminal-only `response.completed` / `response.done` events as retryable empty completions
- reuse bounded retry logic, dropping `previous_response_id` so retries send full context
- return explicit 503 after retry exhaustion instead of successful empty `end_turn`
- preserve text, thinking, tool, web-search, and signature-only reasoning streams
- prove the retry loop stays bounded: a smoke test with an always-empty upstream asserts exactly 1 + 10 attempts, then the explicit 503 (instant via a zero-delay retry test hook, same `*_for_tests` pattern as the websocket pool/continuation helpers)

Scope: this covers the WebSocket live-stream path (`CCP_CODEX_TRANSPORT=websocket`, the default). The buffered HTTP transport translates a terminal-only completion through the reducer and would still return an empty 200; left untouched here since the failure was only ever observed on the live path.

## Failure mode

A live Codex request could end with only `response.completed`. The proxy translated that into HTTP 200 SSE containing `message_start` with `content: []`, followed by `end_turn`. Claude Code then exited mid-turn; its resume repair inserted the synthetic `No response requested.` message, making the failure look like model nonsense.

The new smoke fixture reproduces that exact empty response before the fix and proves a full-context retry afterward.

## TDD evidence

RED:

```text
second response body: event: message_start
... "content":[] ...
... "stop_reason":"end_turn" ...
test result: FAILED. 0 passed; 1 failed
```

GREEN:

```text
smoke_codex_websocket_stream_retries_terminal_only_completion_with_full_context ... ok
```

Assertions cover retry text, three upstream attempts, prior-response use on the failed attempt, and full-context retry without `previous_response_id`.

## Verification

- `cargo fmt --all --check`
- `cargo test --all` — 600 tests passed across unit and integration suites (includes the new exhaustion smoke test, `smoke_codex_websocket_empty_completions_exhaust_to_service_unavailable`, 0.03s)
- `cargo build --release --locked`
- focused Codex regression group — 195 passed
- websocket stream regression group — 5 passed
- CodeRabbit — 0 findings after one TDD-fixed signature-only edge case
- isolated release binary — health passed; combined locally with #68, real Codex request streamed `patched-proxy-ok` and normal `end_turn`

`cargo clippy --all-targets -- -D warnings` remains blocked by 33 pre-existing stable-1.97 warnings outside changed files.

Independent from #68; no merge or deployment performed.

### DIFF
--- a/src/providers/codex/mod.rs
+++ b/src/providers/codex/mod.rs
@@ -20,6 +20,7 @@ use std::sync::Arc;
 
 use crate::anthropic::error::json_error;
 use crate::anthropic::schema::{CountTokensResponse, MessagesRequest};
+use crate::anthropic::sse::parse_sse_events;
 use crate::config;
 use crate::logging::create_logger;
 use crate::monitor::usage_from_anthropic_sse;
@@ -595,7 +596,7 @@ async fn live_stream_response_once(
             generation_started = true;
         }
         append_upstream_sse_payload(&mut upstream_sse_body, &payload);
-        let (chunk, terminal) = match translate_live_stream_payload(&mut translator, &payload, &ctx)
+        let (chunk, terminal) = match translate_live_stream_payload(&mut translator, &payload, None)
         {
             Ok(result) => result,
             Err(message) => {
@@ -651,6 +652,7 @@ async fn live_stream_response_once(
             };
         }
         if translator.has_semantic_output() && !pending_chunk.is_empty() {
+            record_live_stream_downstream_capture(&ctx, &pending_chunk);
             record_live_stream_progress(&ctx, &pending_chunk);
             if terminal {
                 update_continuation_from_upstream(
@@ -684,6 +686,7 @@ async fn live_stream_response_once(
             if pending_chunk.is_empty() {
                 return LiveStreamStart::Response(empty_live_stream_response());
             }
+            record_live_stream_downstream_capture(&ctx, &pending_chunk);
             record_live_stream_progress(&ctx, &pending_chunk);
             return LiveStreamStart::Response(single_live_stream_response(pending_chunk));
         }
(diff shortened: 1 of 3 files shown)


HM-Silvey (approved):
(withdrawn — posted from the wrong account; see review notes in the PR comments)

Aotricx:
Reviewed the full diff (all three commits) plus an independent second-model pass. **Approving.**

**Correctness — verified in code:**
- Retry loop is sound: `drop_live_continuation_for_retry` clears `previous_response_id`/`input_delta` before the bounded backoff path (`MAX_RETRYABLE_LIVE_STREAM_RETRIES = 10`), so retries always resend full context and the loop cannot spin unbounded.
- The discarded empty attempt never calls `update_continuation_from_upstream`, so its response id can't poison the next turn — locked by the smoke assertions on per-attempt `previous_response_id`.
- `pending_chunk` is consumed on every return path (no double-flush possible); first semantic event always emits bytes, so the empty-pending-at-terminal branch is unreachable once output starts.
- Semantic-flag placement covers all five output kinds (text, thinking, tool call, web-search completion, signature-only reasoning), each unit-tested — including the deliberate web-search added-vs-done asymmetry.
- `response.incomplete` deliberately stays non-retryable (a max-tokens truncation would not improve on resend). Sensible.
- Progress-reporting delay from buffering is metrics-only (`monitor.stream_progress`) — no watchdog consumes it, so nothing can time out on buffered silence.

**Test evidence (ran locally on this branch):**
- `cargo fmt --all -- --check` clean
- Full `cargo test`: 600 passed, 0 failed
- Exhaustion smoke test asserts exactly 1 + 10 upstream attempts then the explicit 503 — the retry bound is proven end-to-end, and the zero-delay hook follows the repo's existing `*_for_tests` pattern

**Noted, non-blocking:**
- Scope is the WebSocket live path (the default transport). The buffered path has the same theoretical failure shape; addressed separately in #71.
- The predicate ignores `rate_limits.allowed`; a credited snapshot followed by an upstream abort surfaces the existing missing-terminal error instead of a 429 — correct trade-off, documented in the PR body.

raine:
Thanks

PR #72, opened 2026-07-20
feat: allow a dedicated model for auto-review classifier requests

mulfyx:
## Summary

- detect Claude Code's non-streaming, tool-free autonomous Bash security-review classifier requests
- route only those requests through `CCP_AUTO_REVIEW_MODEL` or the top-level `autoReviewModel` config key
- resolve a direct override model through the normal provider registry while bypassing session affinity for the one-off request
- preserve the session's existing provider affinity after the classifier finishes
- document the setting and cover config precedence, request detection, provider selection, affinity preservation, and end-to-end HTTP routing

## Motivation

Claude Code sends autonomous Bash permission reviews as separate model requests. When the active session uses a large model or high reasoning effort, those small classifier calls use the same expensive route even though a smaller model is sufficient.

This complements the compact-effort specialization added in [#67](https://github.com/raine/claude-code-proxy/pull/67), but applies to a different request class: the Bash auto-review classifier rather than conversation compaction.

## Configuration

Environment:

```sh
CCP_AUTO_REVIEW_MODEL=gpt-5.6-luna
```

`config.json`:

```json
{
  "autoReviewModel": "gpt-5.6-luna"
}
```

Use a direct model ID for deterministic provider selection. When the option is unset, request routing is unchanged.

The override is deliberately narrow: it does not apply to normal messages, streaming requests, tool-using requests, or `/v1/messages/count_tokens`.

## Verification

- `cargo test --locked` — 623 tests passed
- end-to-end HTTP smoke: a classifier request for `gpt-5.6-sol` reached the mock Codex upstream as `gpt-5.6-terra`, while the following normal request still reached it as `gpt-5.6-sol`
- provider-routing coverage for Codex, Grok, Kimi, and Cursor model IDs
- session-affinity regression coverage


raine:
Thanks

PR #80, opened 2026-07-26
fix(nix): vendor Cargo dependencies

jakob1379:
## Summary

- build the package with `rustPlatform.buildRustPackage`
- vendor dependencies from the checked-in `Cargo.lock`

## Why

The custom build phase invoked Cargo without vendored dependencies. Nix's
sandbox blocks network access during builds, so Cargo failed while resolving
`index.crates.io`.

Using the native Rust builder makes the Cargo phase run offline and keeps
dependency versions tied to the lockfile.

## Validation

- `nix build --print-build-logs`
- 667 Rust tests passed during the Nix build
- `./result/bin/claude-code-proxy --version` returned `0.1.25`

## AI usage

OpenAI Codex was used during development to diagnose the failure, implement the
change, and run the validation above.

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -17,7 +17,7 @@ jobs:
       - name: Build package
         run: |
           for attempt in 1 2 3 4 5; do
-            if nix build --no-link --print-build-logs --option sandbox false --option max-jobs 1; then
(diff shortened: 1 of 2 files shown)


raine:
Thanks

PR #83, opened 2026-07-29
fix(codex): honor env proxies for WebSockets

ItsAlbertZhang:
## Summary

Codex HTTP requests already honor standard proxy environment variables, but WebSocket connections were opened directly. This meant the default WSS transport could require OS-level TUN routing even when an HTTP or SOCKS proxy was configured.

This change adds explicit environment-proxy support for Codex WebSocket setup:

- uses `HTTP_PROXY` for `ws://` and `HTTPS_PROXY` for `wss://`
- supports `ALL_PROXY` fallback and `NO_PROXY` bypass
- supports HTTP, HTTPS, SOCKS4/4A, and SOCKS5/5H proxy URLs
- supports proxy credentials where the underlying protocol permits them
- preserves CGI proxy-variable protection
- fails closed instead of silently retrying a selected proxy route directly

It also:

- preserves both platform-native and public WebPKI trust roots
- handles fragmented CONNECT responses and any successful 2xx status
- preserves tunnel bytes read together with the CONNECT response
- keeps proxy authentication out of destination requests, logs, traffic captures, and errors
- strengthens WebSocket handshake validation
- documents non-TUN proxy configuration and restart behavior

## Testing

- `cargo check --locked`
- focused Codex WebSocket proxy integration tests
- full `cargo test --locked` suite with loopback `NO_PROXY` isolation
- strict Clippy with 12 pre-existing baseline lints explicitly allowed

The full test suite passed, including 9 Codex WebSocket proxy integration tests and the existing pooling, continuation, authentication, retry, and smoke-cutover coverage.

Not run: Astro documentation build, because the documentation dependencies were not installed.


raine:
Thanks

PR #86, opened 2026-07-29
fix(codex): report streaming input usage

mulfyx:
## Summary

- populate Codex `message_start.usage.input_tokens` from the existing translated-request token estimate
- apply the behavior to both HTTP and WebSocket streaming
- keep terminal provider usage authoritative in `message_delta`

## Why

Codex streams currently begin with `input_tokens: 0`. Claude Code consumes the
initial usage event for live task and subagent progress, so token usage remains
zero while an agent is running and may never become useful if the stream ends
early.

Anthropic's streaming examples include input usage in `message_start`, while
`message_delta` usage is cumulative. This change therefore seeds only the
initial event and does not emit synthetic intermediate usage deltas that clients
could double-count.

## Verification

- `cargo test --all-targets`
- live Codex SSE: `message_start` reported 11 estimated input tokens; the
  terminal provider event reported the authoritative 12 input / 7 output tokens
- live Claude Code `Agent`: task progress reported 1,406 tokens while running,
  completion reported 1,468 tokens, and the child successfully used `Read` and
  returned the expected fixture marker

## References

- [Anthropic streaming event flow and cumulative usage](https://platform.claude.com/docs/en/build-with-claude/streaming#event-types)
- [Anthropic full stream example with non-zero input usage in `message_start`](https://platform.claude.com/docs/en/build-with-claude/streaming#full-http-stream-response)
- [Headroom issue describing the same zero-input streaming failure](https://github.com/headroomlabs-ai/headroom/issues/1132)
- [Headroom's merged backfill implementation](https://github.com/headroomlabs-ai/headroom/pull/1305)

### DIFF
--- a/src/providers/codex/mod.rs
+++ b/src/providers/codex/mod.rs
@@ -353,10 +353,12 @@ impl Provider for CodexProvider {
         };
 
         if want_stream {
+            let estimated_input_tokens = count_translated_tokens(&translated);
             let sse_bytes = match translate_stream_bytes_with_traffic(
                 &upstream.body,
                 &message_id,
                 model,
+                estimated_input_tokens,
                 ctx.traffic.as_deref(),
             ) {
                 Ok(b) => b,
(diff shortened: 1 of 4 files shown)


raine:
Thanks

PR #88, opened 2026-07-30
fix(codex): limit WebSocket connection churn

### LINKED ISSUE #87: 403: WebSocket upgrade rejected after Codex concurrent connections exceed a certain threshold
After setting the following environment variables:

```powershell
$env:CCP_CODEX_TRANSPORT = "websocket"
$env:CCP_CODEX_PREVIOUS_RESPONSE_ID = "1"
```

new connections eventually fail to establish after the application has been running for some time. The issue is easy to reproduce and occurs particularly frequently in UltraCode mode.

Image

I suspect there may be limits on the OpenAI side regarding either the number of simultaneous WebSocket connections or the number of connection attempts made within a short period.

In any case, the issue no longer occurs after I introduced the following restrictions on WebSocket connections:

* When opening a new connection, if there are already more than 50 active connections, close the 10 least recently active connections, reducing the total to 40.
* Ensure an interval of at least one second between connection attempts.
* When a connection attempt returns a 403 response, wait three seconds and retry once.

I plan to open a related PR later. If you have any feedback or suggestions on how to make the proposed changes more suitable for general use cases, please feel free to let me know.

ItsAlbertZhang:
## Summary

- Prune the reusable Codex WebSocket pool to 40 least-recently-active idle connections when it grows beyond 50.
- Enforce at least one second between fresh WebSocket connection attempts.
- Retry an origin WebSocket upgrade 403 once after a three-second cooldown.
- Keep explicit proxy CONNECT rejections terminal and preserve existing retry and fallback behavior.

## Implementation notes

- Track pool activity independently from the creation timestamp used for TTL expiry.
- Skip leased connections during cleanup and release socket guards before returning reusable connections to the pool.
- Keep origin-versus-proxy connection provenance private to the WebSocket connector.
- Apply the connection timeout separately to each network attempt.

## Testing

- `cargo test --lib providers::codex::websocket::tests::` — 37 passed
- `cargo test --test codex_websocket_proxy` — 1 passed
- `cargo test --test codex_websocket` — 9 passed
- `cargo test --test smoke_cutover smoke_codex_websocket_stream_returns_delta_before_terminal -- --exact` — passed
- `cargo clippy --all-targets --all-features` — passed with pre-existing warnings

Full `cargo test` reaches three pre-existing `cursor_native` mock-upstream HTTP 502 failures unrelated to this change.

Fixes #87

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -47,6 +47,7 @@ crossterm = "0.28"
 jiff = { version = "0.2", default-features = false, features = ["std", "tz-system", "tzdb-zoneinfo"] }
 
 [dev-dependencies]
+tokio = { version = "1", features = ["test-util"] }
 tempfile = "3.14"
 assert_cmd = "2"
 predicates = "3"
--- a/src/providers/codex/websocket.rs
+++ b/src/providers/codex/websocket.rs
@@ -1,5 +1,6 @@
 use std::collections::HashMap;
 use std::pin::Pin;
+use std::sync::atomic::{AtomicU64, Ordering};
 use std::sync::{Arc, Mutex};
 use std::task::{Context, Poll};
 use std::time::{Duration, Instant};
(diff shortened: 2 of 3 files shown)


raine:
Thanks for the thorough fix!

PR #89, opened 2026-07-30
fix(codex): preserve tool choice semantics

mulfyx:
## Summary

- serialize Codex Responses tool-choice modes as the OpenAI wire strings
  `"auto"`, `"none"`, and `"required"` instead of JSON `null`
- translate Anthropic's nested `disable_parallel_tool_use` flag into the
  top-level OpenAI `parallel_tool_calls` field
- keep Responses Lite parallel tool calls disabled
- add wire-level regression coverage for automatic, disabled, required, and
  forced-function tool choices

## Why

Anthropic and OpenAI expose the same controls using different request shapes:

- Anthropic places `disable_parallel_tool_use` inside `tool_choice`.
  With `auto`, it permits at most one tool call; with `any` or a named tool, it
  requires exactly one.
- OpenAI Responses represents `auto`, `none`, and `required` as JSON strings,
  forced functions as `{"type":"function","name":"..."}`, and parallelism as
  the separate top-level `parallel_tool_calls` boolean.

CCP previously represented the three string modes as unit variants inside a
Serde `untagged` enum. Serde serializes those variants as JSON `null`, so
`none` and `any` lost their meaning on the wire. The full Responses lane also
always sent `parallel_tool_calls: true`, ignoring Claude Code's explicit
single-tool constraint.

References:

- [Anthropic: parallel tool use and `disable_parallel_tool_use`](https://platform.claude.com/docs/en/agents-and-tools/tool-use/parallel-tool-use)
- [Anthropic SDK tool-choice types](https://github.com/anthropics/anthropic-sdk-typescript/blob/main/src/resources/messages/messages.ts#L1853-L1912)
- [OpenAI: tool choice](https://developers.openai.com/api/docs/guides/function-calling#tool-choice)
- [OpenAI: parallel function calling](https://developers.openai.com/api/docs/guides/function-calling#parallel-function-calling)
- [OpenAI Responses request types](https://github.com/openai/openai-python/blob/main/src/openai/types/responses/response_create_params.py#L1347-L1349)
- [Codex Responses request shape](https://github.com/openai/codex/blob/main/codex-rs/codex-api/src/common.rs#L1450-L1469)

## Validation

- `cargo test --all-targets`
- formatter and `git diff --check`
- live Codex verification:
  - `tool_choice: none` returned text without a tool call
  - `tool_choice: any` with parallel use disabled returned exactly one tool call
  - a forced named tool called the requested function
  - captured upstream requests contained `"none"`, `"required"`, or the
    function object and the expected `parallel_tool_calls` value

### DIFF
--- a/src/providers/codex/translate/request.rs
+++ b/src/providers/codex/translate/request.rs
@@ -49,11 +49,17 @@ pub enum ServiceTier {
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
-#[serde(untagged)]
-pub enum ResponsesToolChoice {
+#[serde(rename_all = "snake_case")]
+pub enum ResponsesToolChoiceMode {
     Auto,
     None,
     Required,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum ResponsesToolChoice {
+    Mode(ResponsesToolChoiceMode),
     Function {
         r#type: String,
         name: String,
@@ -427,6 +433,7 @@ pub fn translate_request(
     let input = build_input(req);
     let tools = read_tools(req)?;
     let tool_choice = map_tool_choice(req)?;
+    let parallel_tool_calls = !disable_parallel_tool_use(req);
 
     let mut text = ResponsesText {
         verbosity: Some("low".to_string()),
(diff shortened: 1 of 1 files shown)


raine:
Thanks, great as always

PR #94, opened 2026-07-31
fix(codex): fence overlapping compaction attempts

mulfyx:
## Summary

Codex compaction kept one registry entry per Claude session, while asynchronous compaction producers and terminal handlers identified that entry only by session and model.

On base `529e1d6`, a compaction request awaits upstream work before storing its result ([request path](https://github.com/raine/claude-code-proxy/blob/529e1d6da889e4767584e0a6a8574a1ee919f85e/src/providers/codex/mod.rs#L232-L280)). Meanwhile, `store_compaction` replaces the session slot unconditionally, and activation/abort do not identify the originating attempt ([registry and store](https://github.com/raine/claude-code-proxy/blob/529e1d6da889e4767584e0a6a8574a1ee919f85e/src/providers/codex/compaction.rs#L40-L104), [activation and abort](https://github.com/raine/claude-code-proxy/blob/529e1d6da889e4767584e0a6a8574a1ee919f85e/src/providers/codex/compaction.rs#L106-L220)).

With overlapping compactions A and B for the same session, a late A completion could therefore:

- overwrite B's newer native history;
- anchor B's native history with A's portable summary;
- clear B's state after a late A failure.

This change:

- allocates a monotonic `CompactionAttempt` before awaiting upstream compaction;
- publishes a `Preparing` generation immediately;
- permits store, activation, and abort only for the current generation;
- carries the generation through replay and all buffered/live completion and cleanup paths.

The scope is same-session stale-completion/lost-update protection, not a cross-session isolation issue.

## Consistency precedents

These are consistency precedents, not claims that the Responses compaction API mandates generation IDs:

- This repository's continuation state already allocates a turn ID before asynchronous work and rejects stale publish/abort operations ([allocation](https://github.com/raine/claude-code-proxy/blob/529e1d6da889e4767584e0a6a8574a1ee919f85e/src/providers/codex/continuation.rs#L77-L103), [guards](https://github.com/raine/claude-code-proxy/blob/529e1d6da889e4767584e0a6a8574a1ee919f85e/src/providers/codex/continuation.rs#L208-L245)).
- OpenAI documents compacted `encrypted_content` as an opaque item that must be carried into subsequent Responses requests ([compaction guide](https://developers.openai.com/cookbook/examples/gpt-5/codex_prompting_guide#compaction)). Keeping it paired with the portable summary from the same attempt is therefore an application-level consistency requirement.
- OpenAI Codex uses `expectedTurnId` as an optimistic concurrency precondition ([schema](https://github.com/openai/codex/blob/5548c95d66e29aeb994a982db8a378d9453694b0/codex-rs/app-server-protocol/schema/typescript/v2/TurnSteerParams.ts#L6-L10)) and validates it atomically under the active-turn lock ([implementation](https://github.com/openai/codex/blob/5548c95d66e29aeb994a982db8a378d9453694b0/codex-rs/core/src/session/mod.rs#L3977-L4009)).
- RFC 9110 describes conditional writes as protection against parallel lost updates ([If-Match section 13.1.1](https://www.rfc-editor.org/rfc/rfc9110#section-13.1.1)).

## Validation

- The stale-activation regression test fails on the unfenced implementation and passes with this change.
- Deterministic tests cover stale store, activation, abort, replay abort, invalid stale summary, and replay checks while a newer generation is still preparing.
- `cargo test --all-targets`: 828 tests passed.
- `cargo clippy --all-targets -- -D warnings`: passed.
- A controlled runtime A/B race delayed A until B completed; the next turn replayed the newer `native-B` history.

### DIFF
--- a/src/providers/codex/compaction.rs
+++ b/src/providers/codex/compaction.rs
@@ -1,5 +1,6 @@
 use std::collections::HashMap;
 use std::sync::Mutex;
+use std::sync::atomic::{AtomicU64, Ordering};
 use std::time::{SystemTime, UNIX_EPOCH};
 
 use crate::anthropic::sse::parse_sse_events;
@@ -33,15 +34,25 @@ impl std::fmt::Display for CompactionError {
 }
 
 enum CompactionPhase {
+    Preparing,
     Unconfirmed,
     Anchored { portable_summary: String },
 }
 
 struct CompactionState {
+    attempt: CompactionAttempt,
     model: String,
     native_history: Vec<ResponsesInputItem>,
     phase: CompactionPhase,
     updated_at: u64,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct CompactionAttempt(u64);
+
+pub struct CompactionReplay {
+    pub request: ResponsesRequest,
+    pub attempt: CompactionAttempt,
 }
 
 #[derive(Default)]
@@ -51,6 +62,7 @@ struct CompactionRegistry {
 }
 
 static REGISTRY: Mutex<Option<CompactionRegistry>> = Mutex::new(None);
+static NEXT_ATTEMPT_ID: AtomicU64 = AtomicU64::new(1);
 
 pub async fn request_compaction(
     client: &CodexHttpClient,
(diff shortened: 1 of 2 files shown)


raine:
Thanks for the contribution!

Issue #98, opened 2026-08-02
Kimi: multi-part content serializes without a `type` field, breaking subagents and multimodal messages


flaviomartil:
## Summary

The Kimi provider serializes multi-part content without a `type` field, so Kimi rejects the request with `invalid part type: ` (note the empty value after the colon). This affects both tool results and multimodal user messages.

The most visible symptom is that **subagents are unusable**: Claude Code's `Task` tool returns a multi-block `tool_result`, which always trips this path.

## Reproduction

Both against a running `claude-code-proxy serve` with Kimi authenticated.

**1. Multimodal user message**

```sh
curl -s http://127.0.0.1:18765/v1/messages -H 'Content-Type: application/json' -d '{
  "model": "k3", "max_tokens": 16,
  "messages": [{"role": "user", "content": [
    {"type": "text", "text": "what color?"},
    {"type": "image", "source": {"type": "base64", "media_type": "image/png", "data": "<1x1 png>"}}
  ]}]}'
```

```json
{"error":{"message":"the message at position 0 with role 'user' contains an invalid part type: ","type":"invalid_request_error"}}
```

**2. Tool result with more than one part** (what the `Task` tool produces)

Running any Claude Code prompt that spawns a subagent through the proxy:

```
API Error: 502 {"error":{"message":"the message at position 2 with role 'tool' contains an invalid part type: ","type":"invalid_request_error"}}
```

Reproduced on both `k3` and `kimi-for-coding`, so it is not model specific.

## Root cause

`src/providers/kimi/translate/request.rs`:

```rust
#[derive(Debug, Clone, Serialize)]
#[serde(untagged)]
enum KimiUserContentPart {
    Text { text: String },
    ImageUrl { image_url: KimiImageUrl },
}
```

```rust
#[derive(Debug, Clone, Serialize)]
#[serde(untagged)]
enum KimiToolResultPart {
    Text { text: String },
    ImageUrl { image_url: KimiImageUrl },
}
```

`#[serde(untagged)]` drops the discriminant, so these serialize as `{"text": "..."}` and `{"image_url": {...}}`. Kimi expects the OpenAI multimodal shape, `{"type": "text", "text": "..."}` and `{"type": "image_url", "image_url": {...}}`, and reports the missing discriminant as an empty part type.

## Why single-part tool results still work

`tool_result_content` collapses a lone text part back into a plain string before serializing:

```rust
// Collapse to string when only one text part
if parts.len() == 1
[...]

PR #99, opened 2026-08-02
fix(kimi): tag content parts so serde emits the type discriminant

flaviomartil:
Addresses #98.

## Problem

`KimiUserContentPart` and `KimiToolResultPart` are declared `#[serde(untagged)]`, so they serialize without a discriminant:

```json
{"text": "..."}
{"image_url": {"url": "..."}}
```

Kimi expects the OpenAI multimodal shape and rejects the payload:

```
the message at position 0 with role 'user' contains an invalid part type:
the message at position 2 with role 'tool' contains an invalid part type:
```

The value after the colon is empty because the discriminant is missing.

The practical impact is that **subagents do not work through the proxy**. Claude Code's `Task` tool returns a multi-block `tool_result`, which always takes this path. Single-output tools such as Bash and Read are unaffected, because `tool_result_content` collapses a lone text part back into a plain string before serializing.

## Change

Tag both enums:

```rust
#[serde(tag = "type", rename_all = "snake_case")]
```

Output becomes `{"type":"text","text":"..."}` and `{"type":"image_url","image_url":{...}}`.

## Tests

Two regression tests added, one per enum, asserting the discriminant is present:

- `user_content_parts_carry_a_type_discriminant`
- `tool_result_parts_carry_a_type_discriminant`

The existing `translate_tool_result_with_image`, `translate_tool_result_with_unsupported_blocks` and `translate_user_text_and_image_collapse_correctly` assert only on the `text` and `image_url` keys, so they are unaffected.

## Verification

Built and tested against a live Kimi account on macOS arm64. Before the change, a multimodal user message and any subagent invocation both returned `invalid part type`. After the change, both succeed.

### DIFF
--- a/src/providers/kimi/translate/request.rs
+++ b/src/providers/kimi/translate/request.rs
@@ -488,7 +488,7 @@ fn push_user_messages(out: &mut Vec<KimiMessage>, blocks: &[ContentBlock]) {
 }
 
 #[derive(Debug, Clone, Serialize)]
-#[serde(untagged)]
+#[serde(tag = "type", rename_all = "snake_case")]
 enum KimiUserContentPart {
     Text { text: String },
     ImageUrl { image_url: KimiImageUrl },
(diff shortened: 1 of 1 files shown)


raine:
Thanks!

PR #100, opened 2026-08-02
Preserve Codex statusless transport errors

mulfyx:
## Summary

- preserve the descriptive source message for statusless Codex transport errors
- keep structured `detail` values for statusful upstream responses
- cover an HTTP response-body failure after semantic output has already been streamed

## Problem

Codex HTTP body failures use status `0`, a descriptive `message`, and an internal `detail` discriminator such as `http_response_body`. The shared error mapper preferred `detail` whenever it was present, so Claude Code received:

```text
API Error: http_response_body
```

instead of the actionable transport message. This is especially confusing after partial output, where replay is intentionally disabled to avoid duplicating text or tool side effects.

## Fix

For statusless errors, always surface `CodexError.message`. Preserve the existing `detail`-first behavior for statusful upstream errors.

The regression test serves a deliberately truncated HTTP response after a valid text delta and verifies that:

- partial semantic output is preserved
- the stream terminates with `event: error`
- the descriptive response-body transport error is returned
- the internal `http_response_body` discriminator is not exposed as the user-facing message

This follows the same post-commit boundary used by CLIProxyAPI: retry is allowed only before the first deliverable payload; after partial output, the stream is terminated with an error rather than replayed ([bootstrap handler](https://github.com/router-for-me/CLIProxyAPI/blob/41fc5e134631789e98137245f576680c7fb9b322/sdk/api/handlers/handlers_stream.go#L390-L490), [post-payload regression test](https://github.com/router-for-me/CLIProxyAPI/blob/41fc5e134631789e98137245f576680c7fb9b322/sdk/api/handlers/handlers_stream_bootstrap_test.go#L780-L847)). Codex itself also treats a stream ending before `response.completed` as an incomplete attempt ([Codex regression test](https://github.com/openai/codex/blob/main/codex-rs/core/tests/suite/stream_no_completed.rs)).

## Verification

- `cargo fmt --check`
- `cargo test --offline --all-targets`
- `cargo clippy --offline --all-targets -- -D warnings`

All checks pass on the current `main` base (`v0.1.32`).

### DIFF
--- a/src/providers/codex/mod.rs
+++ b/src/providers/codex/mod.rs
@@ -1480,13 +1480,11 @@ fn is_context_window_overflow(message: &str) -> bool {
 }
 
 fn codex_error_message(err: &client::CodexError) -> &str {
-    err.detail.as_deref().unwrap_or({
-        if err.status == 0 {
-            err.message.as_str()
-        } else {
-            "Upstream error"
-        }
-    })
+    if err.status == 0 {
+        err.message.as_str()
+    } else {
+        err.detail.as_deref().unwrap_or("Upstream error")
+    }
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


raine:
Thanks!

PR #105, opened 2026-08-10
Publish the per-request id as a request-id response header

### LINKED ISSUE #104: Responses carry no request-id header, so proxied Claude Code transcripts have no request id
## Description

Responses from the proxy carry no `request-id` header. Claude Code populates the `requestId` field of every transcript record from that header, so records written during a proxied session have no request id at all.

Measured on one proxied session against one native Anthropic session on the same machine:

| transcript | usage-bearing records | with `requestId` |
|---|---:|---:|
| proxied (`gpt-5.6-sol`) | 4 | 0 |
| native Anthropic | 756 | 753 |

A transcript legitimately repeats a record, and consumers de-duplicate those repeats by request id. With the field empty they cannot, and totals double. Claude Code's own `parse_session.py --cache` reported the same 2-request session as:

```
gpt-5.6-sol  (4 turns)
  input total:       90,482   fresh:      47,986   cache_read:       42,496
```

against true figures of 2 turns and 23,993 fresh input tokens. Its `requestId` column shows synthetic `(no-req-L15)` placeholders for every row.

The response body does carry a unique `id` (`msg_<32 hex>`), and for non-streaming responses that value is usable as a substitute key. It is not available in headers for a streaming response, where the id arrives inside the `message_start` event rather than alongside the response.

## Expected Behavior

Every response carries a `request-id` header holding a value that is stable for one upstream request and distinct across requests, for streaming and non-streaming alike, so Claude Code records a usable `requestId` and consumers can de-duplicate transcript records.

## Reproduction Steps

1. Start the proxy: `CCP_ALIAS_PROVIDER=codex claude-code-proxy serve --port 18765 --no-monitor`.
2. POST an Anthropic-shaped request to `http://localhost:18765/v1/messages` with a small `max_tokens` and inspect the response headers: they contain `content-type`, `transfer-encoding` and `date`, with no `request-id`.
3. Run any Claude Code session against the proxy with `ANTHROPIC_BASE_URL=http://localhost:18765`.
4. Inspect the resulting transcript JSONL and observe no record carries a `requestId` field.

## Environment / Context

`claude-code-proxy 0.1.32` (release build, linux-amd64), Claude Code 2.1.226, Linux, codex provider, observed 2026-08-10.

## Discovered During

Investigating why per-session token totals from a proxied Claude Code session were exactly double the true values.

## Suggested Fix

Unverified. `dispatch_request` already mints a per-request `req_id` and threads it through logging and the monitor, and `RequestMonitorGuard` carries it; it simply never reaches the response. Every response path in `src/server.rs` funnels through `monitor_response_body`, which destructures the response parts before the body is streamed, so stamping the header there would cover both streaming and non-streaming responses in one place. Preserving an upstream-supplied `request-id` rather than overwriting it would avoid relabelling a real provider id with a local uuid.

Nitjsefnie:
## Summary

Stamps a `request-id` header on proxy responses so Claude Code records a usable `requestId` in transcripts. Without it, consumers that de-duplicate transcript records by request id count every request twice.

## Related Issues and Pull Requests

Fixes #104

## Changes

- `src/server.rs`: added `REQUEST_ID_HEADER` and stamped it in `monitor_response_body`, using the `req_id` that `dispatch_request` already mints and `RequestMonitorGuard` already carries.
- Applied at that one function rather than at each `return`, because every response path in the module funnels through it, and it is applied to the response parts before the body is streamed, so a streaming SSE response carries the header too.
- An upstream-supplied `request-id` is preserved rather than overwritten.
- `src/server.rs`: added `request_id_header_tests` covering a plain response, a streaming response, an upstream-supplied header, and an error response.

## Testing

`cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo test` all pass on this branch. The four new tests pass.

Verified end to end against a live proxy on Linux with the codex provider:

- Before: response headers were `content-type`, `transfer-encoding`, `date`; a proxied Claude Code session produced 4 usage-bearing transcript records, 0 with `requestId`; `parse_session.py --cache` reported `4 turns` and `fresh: 47,986`.
- After: the response carries `request-id: 288ebaeb-9b8a-4a6a-8ebf-0b6372caa94c`; an equivalent session produced 4 usage-bearing records, 4 with `requestId`, resolving to 2 unique ids; the same parser reported `2 turns` and `fresh: 23,966`, with real ids in its `requestId` column.

The parser was not modified — the id was the only thing missing.

## Follow-ups / Known Limitations

The header is stamped for responses that pass through `monitor_response_body`. Any future response path that bypasses it would need the same treatment.

## Footer

Generated by Claude Opus 5 (implementation, testing, verification)

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -1783,13 +1783,31 @@ async fn dispatch_request(
     response
 }
 
+/// Header Claude Code reads to populate the `requestId` field it writes into
+/// every transcript record.
+pub const REQUEST_ID_HEADER: &str = "request-id";
+
 fn monitor_response_body(response: Response, guard: RequestMonitorGuard) -> Response {
     let status = response.status();
     let outcome = response
         .extensions()
         .get::<NativeResponseOutcome>()
         .cloned();
-    let (parts, body) = response.into_parts();
+    let (mut parts, body) = response.into_parts();
+    // Publish the per-request id this proxy already mints. Anthropic returns
+    // `request-id`, and Claude Code records it as `requestId`; without it every
+    // downstream consumer that de-duplicates transcript records by request id
+    // (its own parser, usage dashboards) counts each request twice, because a
+    // transcript legitimately repeats a record and the id is what resolves it.
+    //
+    // Stamped here rather than at each `return` because every response path in
+    // this module funnels through this function, and it is applied to the parts
+    // before the body is streamed, so a streaming SSE response carries it too.
+    if !parts.headers.contains_key(REQUEST_ID_HEADER)
+        && let Ok(value) = http::HeaderValue::from_str(&guard.req_id)
+    {
+        parts.headers.insert(REQUEST_ID_HEADER, value);
+    }
     let stream = futures_util::stream::unfold(
         (body, guard, outcome),
         move |(mut body, mut guard, outcome)| async move {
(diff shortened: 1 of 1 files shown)


raine:
Cheers

PR #112, opened 2026-08-14
fix Grok web search: 400s on Claude Code's tool declarations, unrenderable hosted-search blocks, and seized tools

### LINKED ISSUE #107: Grok: hosted-search blocks print as "Unsupported content type" in Claude Code
**Environment**

```
claude-code-proxy v0.1.34
Claude Code 2.1.220, VS Code extension
Provider: grok, model grok-4.6
```

**Problem**

The Grok proxy sends a hosted search as two content blocks, `server_tool_use` and `web_search_tool_result`. The Claude Code VS Code extension can not draw these two block types. The extension prints an error line for each block.

---

**Example 1 -- a web search**

*How to reproduce*

1. Point the Claude Code VS Code extension at the proxy. Use model `grok-4.6`.
2. Send: search the web for the top three car rental companies in London

*Observed*

```
Unsupported content type: server_tool_use
Unsupported content type: web_search_tool_result
```

The extension prints two lines for each search. These errors are sometimes reported more than once. The search worked. The answer was correct. Only the display failed.

*Expected*

No error messages.

---

**Note -- Codex provider possibly has the same fault**

A captured Codex session sends the same two block types. This report covers the Grok side only.

### LINKED ISSUE #108: Grok: 400 unsupported tool field: max_uses -- web search never runs
**Environment**

```
claude-code-proxy v0.1.34
Claude Code 2.1.220 (VS Code extension and CLI)
Provider: grok, model grok-4.6
```

**Problem**

The Grok proxy rejects any tool declaration that carries `max_uses`. It returns a 400, and the request never reaches xAI. Claude Code puts `max_uses` on its own `WebSearch` tool, so every session that declares web search hits this.

---

**Example 1 -- a WebSearch tool with a use cap**

*How to reproduce*

```bash
curl -s -X POST http://$PROXY_HOST:$PROXY_PORT/v1/messages \
  -H 'content-type: application/json' -H 'authorization: Bearer unused' \
  -d '{"model":"grok-4.6","max_tokens":64,
       "messages":[{"role":"user","content":"hi"}],
       "tools":[{"name":"WebSearch","description":"Search the web","max_uses":6,
                 "input_schema":{"type":"object",
                   "properties":{"query":{"type":"string"}},
                   "required":["query"]}}]}'
```

*Observed*

```
API Error: 400 unsupported tool field: max_uses
```

The request is rejected. The turn produces no reply.

*Expected*

The request is accepted and the turn replies normally.

### LINKED ISSUE #109: Grok: 400 unsupported tool field: type -- Anthropic's server web_search declaration is rejected
**Environment**

```
claude-code-proxy v0.1.34
Claude Code 2.1.220 (VS Code extension and CLI)
Provider: grok, model grok-4.6
```

**Problem**

The Grok proxy rejects Anthropic's server-tool declaration for web search. It returns a 400, and the request never reaches xAI. Claude Code sends this declaration when the model reaches for web search.

---

**Example 1 -- the server-tool declaration Claude Code sends**

*How to reproduce*

Either ask a Grok session to search the web and let the model reach for search, or send the declaration directly:

```bash
curl -s -X POST http://$PROXY_HOST:$PROXY_PORT/v1/messages \
  -H 'content-type: application/json' -H 'authorization: Bearer unused' \
  -d '{"model":"grok-4.6","max_tokens":64,
       "messages":[{"role":"user","content":"hi"}],
       "tools":[{"type":"web_search_20250305","name":"web_search","max_uses":8}]}'
```

*Observed*

```
API Error: 400 unsupported tool field: type
```

The same `unsupported tool field: type` error repeats on every search attempt. The session then falls back to `WebFetch`. A traffic capture shows four such requests rejected by the proxy, with no upstream call made.

*Expected*

The declaration is accepted. The provider doc (`docs/src/content/docs/providers/grok.md`) says "Claude Code hosted search tools map to Grok-native tools", which is what this declaration asks for.

### LINKED ISSUE #110: Grok: the proxy makes the model run random unsolicited searches
**Environment**

```
claude-code-proxy v0.1.34
Claude Code 2.1.220 (VS Code extension and CLI)
Provider: grok, model grok-4.6
```

**Problem**

The proxy arms every Grok request for search, even a request that declares no tools and sends no system prompt.

The model then searches at random, on turns that need no search.

---

**Example -- a bare `hi` message sometimes runs an X search**

*How to reproduce*

1. Start the proxy and note its host and port.
2. Send this body 25 times. It carries no system prompt and no tools:

```bash
curl -s -X POST http://$PROXY_HOST:$PROXY_PORT/v1/messages \
  -H 'content-type: application/json' -H 'authorization: Bearer unused' \
  -d '{"model":"grok-4.6","max_tokens":1024,
       "messages":[{"role":"user","content":"hi"}]}'
```

3. Count the `server_tool_use` blocks in each response.

*Observed*

In my run, three of the 25 responses ran an X search. The rest answered the greeting. All the responses with searches were identical:

```
server_tool_use        name=x_search   input={"query": "hi"}
x_search_tool_result   content items=0
usage.server_tool_use  {"web_search_requests": 1, "x_search_requests": 1}
```

An X search for the word "hi", zero results, one billed search request, and two content blocks the VS Code extension cannot draw.

The model knew what the message was:

```
"The user said "hi". That's a simple greeting. I should respond in a
 friendly, casual way, as Grok built by xAI."
```

`input_tokens` was 3708 for a two-character message sent with no system prompt.

*Expected*

A greeting is answered. No search runs.

### LINKED ISSUE #111: Grok: asking for a search deletes every other tool the caller declared
**Environment**

```
claude-code-proxy v0.1.34
Claude Code 2.1.220 (VS Code extension and CLI)
Provider: grok, model grok-4.6
```

**Problem**

Ask a Grok session to do a search and every other declared tool is deleted. The proxy replaces the whole tool list with a single hosted search tool, so the model can do nothing except search. Any other work in the same message is dropped.

---

**Example -- one extra sentence deletes `Bash`**

*How to reproduce*

```bash
curl -s -X POST http://$PROXY_HOST:$PROXY_PORT/v1/messages \
  -H 'content-type: application/json' -H 'authorization: Bearer unused' \
  -d '{"model":"grok-4.6","max_tokens":1024,
       "messages":[{"role":"user","content":"MESSAGE"}],
       "tools":[{"name":"Bash","description":"Run a shell command",
                 "input_schema":{"type":"object",
                   "properties":{"command":{"type":"string"}},
                   "required":["command"]}},
                {"name":"WebSearch","description":"Search the web",
                 "input_schema":{"type":"object",
                   "properties":{"query":{"type":"string"}},
                   "required":["query"]}}]}'
```

Run it twice, with `MESSAGE` set to each of these:

```
A:  hi, do pwd in the bash
B:  hi, do pwd in the bash. also search the web for cat pictures
```

Same two tools, same first sentence. The only difference is one extra sentence at the end of B.

*Observed*

```
A   thinking, tool_use
      tool_use     Bash         {"command": "pwd"}                  <- pwd ran

B   thinking, server_tool_use, web_search_tool_result, ..., text
      web_search   {"query": "cat pictures"}                        <- pwd never ran
```

In B the model has no `Bash`. It cannot run the `pwd` I asked for, so only the search half of the message is answered.

*Expected*

`pwd` runs in `Bash` in both A and B. The web search answers the other half of B.

AlexanderKugel:
Fixes #107
Fixes #108
Fixes #109
Fixes #110
Fixes #111

Five defects on the Grok provider, all found by running Claude Code against
the proxy and capturing the traffic. Two of them stop web search working at
all. One commit, 981 tests, docs updated.

### 1. Hosted-search blocks no client draws (#107)

**Symptom.** `Unsupported content type: server_tool_use` and
`Unsupported content type: web_search_tool_result`, two lines per search.

**Cause.** The proxy reports a hosted search as `server_tool_use` plus
`web_search_tool_result`, or an `x_search_tool_result` that is not an Anthropic
type at all. No Anthropic client I can find draws these today.

**Fix.** Report a hosted search as a text block naming the query. Every client
draws text. `CCP_GROK_SEARCH_BLOCKS=native` keeps the previous block types for
a client that does render them.

### 2. `max_uses` failed the request (#108)

**Symptom.** `400 unsupported tool field: max_uses`, no search runs.

**Cause.** Claude Code caps its own `WebSearch` with `max_uses`.
`parse_tools` whitelists five fields and bails on anything else.

**Fix.** Accept it and drop it. That matches the position already documented
for Codex in `reference/compatibility-and-limitations.md` -- the Anthropic cap
is not enforced because the provider exposes no equivalent.

### 3. Anthropic's server `web_search` declaration failed the request (#109)

**Symptom.** `400 unsupported tool field: type`, on every search attempt.

**Cause.** Claude Code escalates to `{"type":"web_search_20250305",
"name":"web_search","max_uses":8}` when the model searches. The declaration
carries no `input_schema`, so it could not take the function path either.

**Fix.** Accept it and map it to Grok hosted `web_search` -- the mapping
`providers/grok.md` already advertised. The match is exact on both type and
name, so a future version, or another tool carrying such a type, is still
rejected.

### 4. Searches on turns that asked for nothing (#110)

**Symptom.** A bare `hi`, sent with no tools and no system prompt, sometimes
answers with an `x_search` for the word "hi". Three times in 25 identical
requests against v0.1.34; zero results, one billed search request, and two
content blocks from #107.

**Cause.** `x_search` was pushed onto every request, including one that
declared no tools, and two directives were appended to the caller's system
prompt telling the model to use the hosted tools.

**Fix.** Gated behind `CCP_GROK_HOSTED_SEARCH`, off by default. By default the
caller's tools and system prompt are left alone. An X-intent turn is still
offered `x_search` as one additional tool, never in place of anything else, so
xAI's index stays reachable -- Claude Code has no X tool of its own.

### 5. Asking for a search deleted the caller's other tools (#111)

**Symptom.** With `Bash` and `WebSearch` declared, `hi, do pwd in the bash`
runs `pwd`. Add `also search the web for cat pictures` and `Bash` is gone from
the request, so `pwd` never runs. Reproduced five times out of five.

**Cause.** A turn matching one of fourteen substrings had its whole `tools`
array replaced with a single hosted tool and `tool_choice: required`. Nine
substrings route to `x_search`, five to `web_search`. Matching is
`text.contains`, so "uni**x search** for files" triggers it too.

**Fix.** Same gate as #110. On the default path nothing is removed from
the tool list and `tool_choice` stays whatever the caller sent.

#110 and #111 are the changes here that alter behaviour you chose
deliberately, so they are the ones to push back on if any. Commits 259f2cc and
7ad36f7 say the forcing exists so shell tools cannot become a search fallback,
which is a real problem -- I have kept the whole mechanism intact behind
`CCP_GROK_HOSTED_SEARCH=1`, which restores the old behaviour exactly.

### Verification

`cargo fmt --check`, `clippy -D warnings`, `build --all`, `test --all` all
pass. 962 tests before, 981 after: 19 new, none removed, none ignored.

Verified end to end against live Grok, on this commit:

- Anthropic's server `web_search` declaration carrying `max_uses` returns 200.
  The search runs, and the response contains no `server_tool_use` and no
  `web_search_tool_result` -- the queries arrive as text.
- `hi, do pwd in the bash. also search the web for cat pictures`, with `Bash`
  and `WebSearch` declared, calls both: `Bash` for `pwd` and the caller's own
  `WebSearch` for the search. Both halves of the message are answered.
- A bare `hi` with no tools answers the greeting and runs no search.

Every defect above was reproduced against the released v0.1.34 binary first,
so these are fixes rather than regressions.

`providers/grok.md` and `reference/configuration.md` are updated -- that
provider page described the old search behaviour as current. `CHANGELOG.md` is
untouched.

Happy to split this into separate PRs if you'd prefer.

### DIFF
--- a/docs/src/content/docs/providers/grok.md
+++ b/docs/src/content/docs/providers/grok.md
@@ -38,20 +38,20 @@ The proxy translates Claude messages, function tools, tool results, thinking con
 
 Search reaches Grok-native tools when the caller asks for it:
 
-- A caller that declares Anthropic's hosted `web_search` tool gets Grok hosted
-  web search.
-- A caller that runs its own search tool keeps it. The proxy does not replace a
-  client-side search tool with a hosted one.
+- Anthropic's `web_search_20250305` declaration maps to Grok hosted web search.
+  The Grok CLI endpoint accepts the minimal declaration without domain or
+  location constraints.
+- A caller-managed search tool remains a function tool for the caller to run.
 - An X or Twitter query is additionally offered hosted `x_search`, which the
-  model can use or ignore. It never replaces the caller's other tools.
+  model can use or ignore alongside the caller's tools.
 - Citations and search usage return in Anthropic-compatible usage fields.
-- `CCP_GROK_HOSTED_SEARCH=1` restores the previous behaviour, in which hosted
-  tools replace the caller's search tools and an explicit search turn is forced.
+- `CCP_GROK_HOSTED_SEARCH=1` enables a policy where hosted tools replace caller
+  search tools and explicit search turns require a tool call.
 
-A hosted search is reported as a text block naming the query, which every
-Anthropic client can draw. `CCP_GROK_SEARCH_BLOCKS=native` reports it as
-`server_tool_use` plus `web_search_tool_result` or `x_search_tool_result`
-instead, for a client that renders those block types.
+A hosted search is reported as a text block naming the query.
+`CCP_GROK_SEARCH_BLOCKS=native` preserves `server_tool_use` plus
+`web_search_tool_result` or `x_search_tool_result` for clients that consume
+hosted-tool blocks.
 
 ## OpenAI-compatible APIs
 
@@ -81,11 +81,16 @@ Traffic captures redact Anthropic image data and upstream image data URLs.
 - `CCP_GROK_BASE_URL` or `grok.baseUrl` changes the API base URL.
 - `CCP_GROK_CLIENT_VERSION` or `grok.clientVersion` changes the client version header.
 - `CCP_GROK_TOOL_IMAGE` selects `omit`, `reattach`, `inline`, or `reject`.
-- `CCP_GROK_HOSTED_SEARCH` restores hosted search tools replacing the caller's own.
+- `CCP_GROK_HOSTED_SEARCH` enables hosted search replacement and forcing.
 - `CCP_GROK_SEARCH_BLOCKS` selects `text` or `native` hosted-search reporting.
 
 See [Configuration](/reference/configuration/) for defaults.
 
 ## Limitations and troubleshooting
+
+The Grok CLI hosted web-search endpoint has no equivalent for Anthropic's
+`max_uses`, domain filters, or user location. The proxy accepts and omits a
+valid `max_uses` value, while non-null domain and location constraints return a
+request error rather than weakening the requested search scope.
 
 A successful login does not guarantee every model is enabled for the account or region. Model rejection and upstream errors are surfaced to Claude Code. Use `grok auth status` for token state, inspect the failed request in the monitor, and use the structured log or error capture for the full redacted response.
--- a/docs/src/content/docs/providers/grok.md
+++ b/docs/src/content/docs/providers/grok.md
@@ -36,11 +36,22 @@ ANTHROPIC_SMALL_FAST_MODEL=grok-4.6 \
 
 The proxy translates Claude messages, function tools, tool results, thinking controls, token usage, and streaming events. Grok reasoning text appears as Claude Code thinking blocks.
 
-Claude Code hosted search tools map to Grok-native tools:
+Search reaches Grok-native tools when the caller asks for it:
 
-- General web queries use hosted web search.
-- X queries use hosted `x_search`.
-- Citations and search usage return in Anthropic-compatible content and usage fields.
+- A caller that declares Anthropic's hosted `web_search` tool gets Grok hosted
+  web search.
+- A caller that runs its own search tool keeps it. The proxy does not replace a
+  client-side search tool with a hosted one.
+- An X or Twitter query is additionally offered hosted `x_search`, which the
+  model can use or ignore. It never replaces the caller's other tools.
+- Citations and search usage return in Anthropic-compatible usage fields.
+- `CCP_GROK_HOSTED_SEARCH=1` restores the previous behaviour, in which hosted
+  tools replace the caller's search tools and an explicit search turn is forced.
+
+A hosted search is reported as a text block naming the query, which every
+Anthropic client can draw. `CCP_GROK_SEARCH_BLOCKS=native` reports it as
+`server_tool_use` plus `web_search_tool_result` or `x_search_tool_result`
+instead, for a client that renders those block types.
 
 ## OpenAI-compatible APIs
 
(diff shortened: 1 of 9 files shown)


raine:
Thanks

PR #122, opened 2026-08-25
codex: treat an empty successful completion after a tool_result tail as a valid end_turn

dmitryanchikov:
## Summary

When a conversation ends `assistant: tool_use` → `user: tool_result` and the tool call itself
delivered the user-visible answer (e.g. a messaging/reply tool), gpt-5.x models legitimately
end the follow-up turn with **zero output items** — a successful terminal event with nothing
to add. The empty-completion protection from #70/#71 treats this as a failure: it retries
(bounded 1+10), burns ~3 minutes, then surfaces `503 Codex completed without producing
output` — and the Anthropic-side client (Claude Code) retries the whole request on top,
producing multi-minute hangs on trivial turns.

For a Telegram channel bot whose reply tool IS the answer, this tail shape occurs on
essentially every turn: **~68% of its real turns failed this way over 2h** (2026-08-25,
websocket transport, main @ 0185409).

## Deterministic repro

```json
{"model":"gpt-5.5","max_tokens":1024,"stream":false,
 "system":"You are a channel agent. When a user pings you, reply via the reply tool. The tool call IS your answer to the user.",
 "tools":[{"name":"reply","description":"Send a reply message to the chat. This delivers your answer to the user.","input_schema":{"type":"object","properties":{"text":{"type":"string"}},"required":["text"]}}],
 "messages":[
  {"role":"user","content":"ping"},
  {"role":"assistant","content":[{"type":"tool_use","id":"toolu_01","name":"reply","input":{"text":"pong"}}]},
  {"role":"user","content":[{"type":"tool_result","tool_use_id":"toolu_01","content":"Message sent."}]}]}
```

Before: `503` after **175s**, empty. (Adding a system line "always end with a short text
line" makes the identical request return 200 in 1.8s — the model is fine; the empties are
semantic, not transient.)

## Change

Both empty-completion sites (buffered and live-stream) now consult the translated request's
tail: a `function_call_output` tail means the empty completion is the turn's real end, and it
translates into a valid empty `end_turn` instead of entering the retry loop. Empty
completions **without** a tool-result tail keep the full #70/#71 retry-then-503 behavior.

## After (verified in production, both transports)

- non-streaming: `200` in ~1.7s, `{"content":[],"stop_reason":"end_turn","role":"assistant",...}`
- streaming: `200` in ~1.6s, well-formed SSE (`message_start` → `content_block_start/stop` →
  `message_delta` → `message_stop`) — accepted by Claude Code.

Full `cargo test` green. Happy to add a regression test in the #71 harness style if you can
point me at the preferred fixture shape.

### DIFF
--- a/src/providers/codex/mod.rs
+++ b/src/providers/codex/mod.rs
@@ -334,6 +334,7 @@ impl CodexProvider {
                     attempt: compaction_attempt,
                 },
                 configured_transport,
+                tail_is_successful_tool_result(&body),
             )
             .await;
             log.info(
@@ -357,6 +358,7 @@ impl CodexProvider {
         let request_continuation = continuation.clone();
         let mut continuation = Some(continuation);
         let mut attempt = 0_u32;
+        let successful_tool_tail = tail_is_successful_tool_result(&body);
         let upstream = loop {
             let response = match client
                 .post_codex_for_owner(&translated, &ctx, continuation.as_ref())
@@ -387,6 +389,12 @@ impl CodexProvider {
                 }
             };
             if !is_empty_codex_success_completion(&response.body) {
+                break response;
+            }
+            if accept_empty_completion(successful_tool_tail, attempt) {
+                // Second consecutive empty after a successful tool_result tail: the
+                // empty IS the turn's end — let it translate into a valid empty
+                // end_turn (see accept_empty_completion).
                 break response;
             }
             // A successful terminal event with no output would translate into
@@ -714,9 +722,14 @@ async fn live_stream_response(
     continuation: ContinuationReservation,
     compaction: LiveStreamCompaction,
     transport: config::CodexTransport,
+    successful_tool_tail: bool,
 ) -> Response {
     let model = model.to_string();
     let request_continuation = continuation.clone();
+    // Counts consecutive empty-completion retries specifically (not connection retries):
+    // accept_empty_completion() flips after the FIRST one, so the second consecutive
+    // empty is delivered as a valid empty end_turn instead of another retry.
+    let mut empty_completion_attempts = 0_u32;
     let mut cleanup = LiveRequestStateCleanup::new(
         request_continuation.clone(),
         ctx.session_id.clone(),
@@ -786,6 +799,7 @@ async fn live_stream_response(
             request_continuation.clone(),
             request_body.clone(),
             compaction,
+            accept_empty_completion(successful_tool_tail, empty_completion_attempts),
         )
         .await
         {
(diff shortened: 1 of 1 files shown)


dmitryanchikov:
Force-pushed a v2 after internal review flagged the original as too broad — the empty-acceptance is now doubly gated:

1. **Predicate narrowed**: runs on the original Anthropic body and requires a `tool_result` **without `is_error`** — an error result demands a model reaction, so a deterministic empty there stays a loud failure.
2. **One retry as the glitch discriminator**: a qualifying tail alone isn't proof (an internal tool result has the same shape and can meet a transient upstream glitch, which is what #70/#71 exist for). The first empty still retries once; only the **second consecutive** empty is accepted as a valid empty `end_turn`. Both transports share one `accept_empty_completion()`.

Added tests: tail predicate (success/error/non-tool/ordering/empty), the accept truth table, and a live-path behavioral pair (flag off → Retry with the empty-completion detail; flag on → delivered stream ending in `message_stop`). Verified in production: the repro returns 200 in ~4.5s (one discriminator roundtrip included), and an `is_error`-tail control produced an actual model reaction (thinking + a retry tool call) rather than a silent empty.

pierzchalski:
I reproduced this with Claude Code 2.1.278 and a proxy built from 55bf0b58 (v0.1.35 + 6 commits), using `CCP_CODEX_TRANSPORT=http`. Claude Code’s Agent-tool subagents deliver their report with a `SubagentHandback` tool call and are instructed to “call it and then stop,” so the request after that tool result is exactly this tail and the model legitimately returns nothing. In one session, 15 handbacks succeeded across 7 `gpt-5.6-sol` and `gpt-5.6-terra` subagents; 8 then stalled for 19–56 minutes, with 5 ending in `503 Codex completed without producing output` and 3 interrupted by the user first, while the other 7 happened to emit a follow-up text turn and completed normally. Claude Code showed each stalled subagent as still running, then reported “Agent terminated early due to an API error” despite the report already being delivered, and each stall re-sent the full subagent context up to 10 times. `f7ee7e8` cherry-picks cleanly onto v0.1.41 (`2b639129`); we are deploying it locally.

pierzchalski:
We captured Claude Code 2.1.278’s requests after a `SubagentHandback` tool result via a logging shim between Claude Code and the proxy. Their tails are `user[tool_result] → system[text]` and `user[tool_result, text] → system[text]`: Claude Code appends PostToolUse hook output as trailing `text` blocks in the tool-result message and as a trailing `system`-role message in `messages[]`. Since `tail_is_successful_tool_result` requires the non-error tool result to be the final block of the final message, acceptance never fires, so the 10 retries and 503 remain. We’re carrying a local patch that skips both trailing forms after the last tool result, with unit tests for each; happy to open it against this branch if wanted. Without hooks, the tail is exactly what this PR expects.

Issue #124, opened 2026-08-28
Test `cancellation_while_replacement_startup_is_blocked_aborts_request_state` fails: replacement startup never reaches the blocked handshake


AlexanderKugel:
**Environment**

```text
commit: 0185409 (upstream main, "compare function_call arguments semantically ...", #119)
rustc 1.97.1 (8bab26f4f 2026-07-14)
cargo 1.97.1 (c980f4866 2026-06-30)
Linux 6.18.33.2-microsoft-standard-WSL2
```

**Problem**

One library test fails.

**How to reproduce**

```bash
cargo test --lib cancellation_while_replacement_startup_is_blocked_aborts_request_state
```

**Observed result**

```text
running 1 test
test providers::codex::tests::cancellation_while_replacement_startup_is_blocked_aborts_request_state ... FAILED

thread 'providers::codex::tests::cancellation_while_replacement_startup_is_blocked_aborts_request_state'
panicked at src/providers/codex/mod.rs:2532:14:
replacement startup did not reach the blocked handshake: Elapsed(())

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 864 filtered out; finished in 2.07s
```

**Expected**

The test passes.

**Notes**

The failure repeated on every run, including at commit `0185409` with no local change.

The assertion is a 2-second timeout, and it was only tested on WSL2. It may pass on a different machine. 

PS. just solved it my increasing the timeout.

PR #125, opened 2026-08-28
fix(grok): stop rejecting hosted web search domain and location filters

### LINKED ISSUE #123: Grok hosted web search constraints cause a proxy 400, after which Claude Code retries without them
**Environment**

```text
claude-code-proxy live grok-claude instance, 127.0.0.1:18765
Claude Code 2.1.220, VS Code extension
provider: grok
model: grok-4.6
Grok client version (proxy header): 0.2.93
```

**Problem**

Claude Code can declare Anthropic hosted web search with `allowed_domains`, `blocked_domains`, or `user_location`. Grok's hosted `web_search` does not have equivalent fields.

When any of those fields is non-null, claude-code-proxy returns HTTP 400 without calling the Grok agent:

```text
API Error: 400 Grok hosted web search does not support allowed_domains
```

The design is meant to stop incompatible requests from reaching Grok. Claude Code, however, receives the proxy's HTTP 400, presents that error to the model as an error `tool_result`, and retries the search without `allowed_domains`. The retry is accepted and runs as an unrestricted web search.

The hard HTTP 400 therefore adds another turn, tokens, and latency, while the search still runs without the requested restriction. The intended protection from the 400 is not achieved.

---

**How to reproduce**

Reliable replay against `127.0.0.1:18765`:

```bash
curl -s -X POST 'http://127.0.0.1:18765/v1/messages?beta=true' \
  -H 'content-type: application/json' \
  -H 'anthropic-version: 2023-06-01' \
  -H 'authorization: Bearer unused' \
  -d '{
    "model": "grok-4.6",
    "max_tokens": 64,
    "stream": false,
    "messages": [{"role": "user", "content": "What are LangChain Deep Agents LangGraph 2026 tools"}],
    "tools": [{
      "type": "web_search_20250305",
      "name": "web_search",
      "max_uses": 8,
      "allowed_domains": ["langchain.com", "docs.langchain.com", "blog.langchain.com"]
    }]
  }'
```

**Observed results**

1. The proxy rejects the hosted search containing `allowed_domains`. Status: 400. No upstream Grok call is made.
2. Claude Code presents the proxy error as a `tool_result` for the original `WebSearch` tool use, with `is_error: true`.
3. On the next assistant turn, the model issues two new `WebSearch` calls for the same topic without `allowed_domains`.
4. Those calls return search results.

**Expected**

A site-limited or location-limited hosted search should not consume a failed round only to be retried without its constraint on the next turn.

If the provider cannot enforce the constraint, the proxy should still be able to complete the search in one round. The existing hard 400 behavior should remain available as an opt-in legacy mode rather than being the default.

When the proxy does reject a request, the error should name every unsupported setting in that request -- `allowed_domains`, `blocked_domains`, `user_location`, or any combination of them -- instead of reporting only the first field it finds.

---

**Proposal**

```text
CCP_SEARCH_CONSTRAINTS=hard|soft|warning
```

Default: `soft`.

| Value | Behaviour |
| --- | --- |
| `hard` | Preserve the current proxy 400. In the observed Claude Code flow, the model then retries without the constraint. End effect: unrestricted search in two rounds. Legacy opt-in. |
| `soft` | Drop the unsupported fields, turn the constraints into instructions to the model, and send the hosted search. Best effort only. There is no guarantee that the model stays within the named sites. In testing, every listed source stayed in-domain. One round instead of two, saving tokens and latency. |
| `warning` | Drop the unsupported fields, log a warning, and send the hosted search without a prompt hint. This uses the same unrestricted search configuration as the retry after `hard`, but avoids the failed round and saves tokens and latency. |

In the observed retry path, `hard` and `warning` both lead to an unrestricted hosted-search request. `warning` avoids the failed round. `soft` is the only mode that still attempts to preserve the caller's requested sites.

The setting applies to providers that cannot enforce these Anthropic hosted-search options. Grok is the first such provider. The documentation should list the providers to which the setting applies.

`max_uses` remains dropped, as it is today.

---

**Why `soft` is the default**

Grok cannot enforce `allowed_domains`. With `hard`, Claude Code retries without domain restrictions, which does not preserve the original intent. With `soft`, the proxy turns each unsupported field into a direct instruction and appends it to the `instructions` field of the upstream Grok request. That field carries the system prompt.

This was tested 10 times with `soft` and domain restrictions, and 10 times without a domain list (`warning` / unrestricted). Both groups used the same query.

| Group | Runs | Total [inside, outside] | In-domain share |
| --- | --- | --- | --- |
| soft (directive) | 10/10 | [161, 0] | 100.0% |
| open (no domains) | 10/10 | [147, 210] | 41.2% |

Every `soft` run listed in-domain sources only. Grok applied the instruction by adding `site:` filters to its own search queries. The unrestricted group returned more off-domain sources than in-domain ones.

The open group reaches 41.2% because the query names LangChain, so an unrestricted search finds the allowed domains anyway. This is why both groups must share one query.

`soft` remains a bias, not an enforcement mechanism. Grok can ignore the instruction. The measurement shows that it followed the instruction in all 10 runs.

**Status:** A fix is implemented and running in the fork `AlexanderKugel/claude-code-proxy`. The default is `soft`. Set `CCP_SEARCH_CONSTRAINTS=hard` to keep the current 400 behavior. A pull request to this repository will follow.

AlexanderKugel:
Fixes #123.

## Problem

Claude Code can declare Anthropic hosted web search with `allowed_domains`, `blocked_domains`, or `user_location`. Grok's hosted `web_search` has no equivalent fields, so the proxy returns HTTP 400 without calling Grok.

Claude Code surfaces that 400 to the model as an error `tool_result`. The model then retries the same search without the filter, and the retry is accepted and runs unrestricted. The rejection therefore costs an extra round and the caller's constraint is lost anyway.

## Change

`CCP_SEARCH_CONSTRAINTS` selects the behaviour for providers that cannot enforce these fields:

| Value | Behaviour |
| --- | --- |
| `soft` (default) | Drop the fields and turn each one into an instruction appended to the upstream request's `instructions` field. |
| `warning` | Drop the fields, log a warning, send the search unrestricted. |
| `hard` | Keep the current HTTP 400. |

`hard` preserves today's behaviour for anyone who wants it. The 400 message now names every unsupported field in the request instead of only the first one found.

`max_uses` stays dropped, as before.

Example of the appended instruction:

```text
You are only allowed to search allowed_domains=["langchain.com","docs.langchain.com","blog.langchain.com"].
```

Values are copied verbatim from the caller. The proxy does not parse them.

## Why `soft` is the default

Measured with 10 live searches per group, same query, three allowed domains:

| Group | Runs | Total [in-domain, off-domain] | In-domain share |
| --- | --- | --- | --- |
| `soft` | 10/10 | [161, 0] | 100.0% |
| no domain list | 10/10 | [147, 210] | 41.2% |

Grok applies the instruction by adding `site:` filters to its own search queries. `soft` is a bias, not enforcement; the model can ignore it.

## Notes

- Unit tests cover all three modes, the multi-field error message, and the verbatim copying of `user_location`.
- No version bump and no changelog entry, as those are yours to make.
- `cargo fmt --check` and `cargo clippy --lib -- -D warnings` are clean. The library suite passes except `providers::codex::tests::cancellation_while_replacement_startup_is_blocked_aborts_request_state`, which fails on `main` without this change and is reported in #124.

### DIFF
--- a/docs/src/content/docs/providers/grok.md
+++ b/docs/src/content/docs/providers/grok.md
@@ -40,7 +40,9 @@ Search reaches Grok-native tools when the caller asks for it:
 
 - Anthropic's `web_search_20250305` declaration maps to Grok hosted web search.
   The Grok CLI endpoint accepts the minimal declaration without domain or
-  location constraints.
+  location constraints. `CCP_SEARCH_CONSTRAINTS` selects what happens when
+  Claude Code still sends those fields: `soft` (default) copies constraints into a
+  prompt hint, `warning` drops them and logs, `hard` returns 400.
 - A caller-managed search tool remains a function tool for the caller to run.
 - An X or Twitter query is additionally offered hosted `x_search`, which the
   model can use or ignore alongside the caller's tools.
@@ -83,6 +85,8 @@ Traffic captures redact Anthropic image data and upstream image data URLs.
 - `CCP_GROK_TOOL_IMAGE` selects `omit`, `reattach`, `inline`, or `reject`.
 - `CCP_GROK_HOSTED_SEARCH` enables hosted search replacement and forcing.
 - `CCP_GROK_SEARCH_BLOCKS` selects `text` or `native` hosted-search reporting.
+- `CCP_SEARCH_CONSTRAINTS` selects `soft`, `warning`, or `hard` for Anthropic
+  hosted-search domain and location fields Grok cannot enforce.
 
 See [Configuration](/reference/configuration/) for defaults.
 
--- a/docs/src/content/docs/reference/compatibility-and-limitations.md
+++ b/docs/src/content/docs/reference/compatibility-and-limitations.md
@@ -69,8 +69,9 @@ claude-code-proxy targets Claude Code's practical Anthropic API usage rather tha
 - Model availability varies by account and region.
 - Hosted general web search and X search are translated with citations and usage.
 - Hosted web search omits `max_uses` because the Grok CLI endpoint exposes no
-  equivalent cap. Non-null domain filters and user location are rejected because
-  dropping them would weaken the caller's requested search scope.
+  equivalent cap. Non-null domain filters and user location follow
+  `CCP_SEARCH_CONSTRAINTS` (`soft` default, `warning`, or `hard`) because the
+  endpoint cannot enforce them.
 - The implemented multimodal path does not claim general image or video compatibility.
 
 ## Cursor Agent
--- a/docs/src/content/docs/reference/configuration.md
+++ b/docs/src/content/docs/reference/configuration.md
@@ -70,6 +70,7 @@ All keys are optional. An unreadable file, malformed JSON, or incompatible field
 | `CCP_LOG_VERBOSE` | `log.verbose` | `false` | Preserves full string fields in structured logs when present, regardless of its value. |
 | `CCP_TRAFFIC_LOG` | none | `false` | Enables full request captures for `1`, `true`, or `yes`. |
 | `XDG_STATE_HOME` | none | `~/.local/state` | State base on macOS and Linux. |
+| `CCP_SEARCH_CONSTRAINTS` | none | `soft` | How to treat Anthropic hosted-search options a provider cannot enforce (`allowed_domains`, `blocked_domains`, `user_location`). `soft` drops them and copies constraints into a prompt hint. `warning` drops them and logs. `hard` returns 400. First provider: Grok. Codex maps domain filters natively and ignores this setting. |
 
 `CCP_CONFIG_DIR` affects `config.json` and file-backed provider auth. It does not relocate the state directory.
 
--- a/src/config.rs
+++ b/src/config.rs
@@ -507,6 +507,40 @@ pub fn grok_search_blocks() -> GrokSearchBlocks {
     parse_grok_search_blocks(std::env::var("CCP_GROK_SEARCH_BLOCKS").ok().as_deref())
 }
 
+// ---------------------------------------------------------------------------
+// Hosted-search constraints that a provider cannot enforce
+// (CCP_SEARCH_CONSTRAINTS)
+// ---------------------------------------------------------------------------
+
+/// How the proxy treats Anthropic hosted-search options that the upstream
+/// provider cannot enforce (`allowed_domains`, `blocked_domains`,
+/// `user_location`).
+///
+/// Applies to providers that lack those fields. First provider: Grok. Codex
+/// maps domain filters natively and does not use this policy.
+///
+/// `Soft` is the default: drop the fields and copy constraints into a prompt hint.
+/// `Warning` drops them, logs, and continues with no hint. `Hard` is the
+/// legacy 400. Unknown values fall back to `Soft`.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum SearchConstraints {
+    Soft,
+    Warning,
+    Hard,
+}
+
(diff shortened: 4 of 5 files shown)


AlexanderKugel:
Merged current raine/main into this branch so CI picks up the nixpkgs crates.io 403 fix (`ec5baab`). The PR change is unchanged. Full checks already passed; flake should pass on this push.

oh-yeah-sea-kit2:
Data point on the premise: Grok's hosted `web_search` does accept domain filters on the endpoint this proxy talks to.

I ran into #123, wrote a local patch, and only then found this PR. While validating the patch I sent requests straight to `https://cli-chat-proxy.grok.com/v1/responses` (the `DEFAULT_BASE_URL` in `src/providers/grok/client.rs`), with the same headers `client.rs` sends and a SuperGrok OAuth token. The tool shape xAI documents for web search is:

```json
{"type": "web_search", "filters": {"allowed_domains": ["blog.rust-lang.org"]}}
```

Three live runs, `grok-4.6`, streaming, same kind of query:

| Tool | Search-result sources | Citations (`url_citation`) |
| --- | --- | --- |
| `{"type":"web_search"}` (control) | mixed hosts | en.wikipedia.org x4, www.rust-lang.org x4 |
| `filters.allowed_domains=["blog.rust-lang.org"]` | 39/39 blog.rust-lang.org | 3/3 blog.rust-lang.org |
| `filters.excluded_domains=["wikipedia.org"]` | no wikipedia among the results | 0 wikipedia (control had 4) |

All three returned 200. The bare `wikipedia.org` entry also covered the `en.wikipedia.org` subdomain.

Constraints documented at docs.x.ai: at most 5 domains per list, and `allowed_domains` / `excluded_domains` cannot both be set in one request.

If that reproduces for you, the filters could be mapped natively for Grok, the way `src/providers/codex/search.rs` already maps `SearchFilters { allowed_domains, blocked_domains }` for Codex. That makes the caller's constraint enforced upstream rather than a bias the model can ignore, which your own measurement notes ("`soft` is a bias, not enforcement").

`CCP_SEARCH_CONSTRAINTS` would still earn its place: `user_location` has no Grok equivalent, and the over-5-domains and both-lists-set cases still need a policy.

Happy to post the exact request bodies and raw SSE if that helps.

AlexanderKugel:
> Constraints documented at docs.x.ai: at most 5 domains per list, and `allowed_domains` / `excluded_domains` cannot both be set in one request.
> 
> If that reproduces for you, the filters could be mapped natively for Grok, the way `src/providers/codex/search.rs` already maps `SearchFilters { allowed_domains, blocked_domains }` for Codex. That makes the caller's constraint enforced upstream rather than a bias the model can ignore, which your own measurement notes ("`soft` is a bias, not enforcement").

Good catch. Native `filters` mapping is the right next step, same idea as Codex `SearchFilters`.
You probably know that, but just in case, the `soft` workaround does the job today, so you can use it. It's not the cleanest solution, but it's better than nothing :)
>
> `CCP_SEARCH_CONSTRAINTS` would still earn its place: `user_location` has no Grok equivalent, and the over-5-domains and both-lists-set cases still need a policy.
>

After native domain mapping, `CCP_SEARCH_CONSTRAINTS` would still cover:
- `user_location`
- more than 5 domains on one list

`max_uses` stays dropped, as today. I will have a look.

AlexanderKugel:
Superseded by #161. That PR maps Anthropic hosted `web_search` onto Grok natively (`filters` and nested `user_location`), which is the approach @oh-yeah-sea-kit2 pointed out here.

AlexanderKugel:
Closing in favor of #161.

PR #129, opened 2026-09-04
feat(codex): allow gpt-6-astra

olzn:
Adds `gpt-6-astra` (released 3 September) to the Codex provider.

- `ALLOWED_MODELS` (`translate/model_allowlist.rs`): selectable by name and via the `-fast` priority-tier alias.
- `uses_responses_lite`: it serves from the Responses Lite lane alongside the gpt-5.6 tier. Verified with a ChatGPT-authenticated Codex account: a request through the proxy returns 200 with `stop_reason: end_turn`, and a full Claude Code turn with `ANTHROPIC_MODEL=gpt-6-astra` completes.
- `CODEX_MODELS` (`registry.rs`): so the model and its `-fast` sibling appear in `GET /v1/models`.

Not changed: `MODEL_ALIASES`. The Claude aliases keep pointing at the gpt-5.6 tier; anyone who wants opus or fable on Astra can set `ANTHROPIC_DEFAULT_OPUS_MODEL=gpt-6-astra` on the Claude Code side. `full_lane_web_search_model` is also untouched, as Astra is not lite-only.

Via Codex the model reports a 272k context window (872k max), so the 1M hint does not apply.

`cargo fmt --check` clean. `cargo test --lib`: 853 of 854 pass in the full parallel run; the one failure, `cancellation_while_replacement_startup_is_blocked_aborts_request_state`, is a timing test that passes in isolation on both this branch and unpatched v0.1.35.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #130, opened 2026-09-06
feat(codex): extend the model allowlist from the Codex CLI's model cache

GiulianoCosta71:
## Why

Every new Codex model currently needs a code change plus a release before the proxy accepts it (#129 for `gpt-6-astra`, #113/#114 for `gpt-daybreak-blue`). Meanwhile the Codex CLI already knows the model: it keeps `~/.codex/models_cache.json` fresh from the server, including `supported_in_api`, `use_responses_lite` and `visibility` per model.

## What

Read that cache at runtime (`$CODEX_HOME/models_cache.json`, default `~/.codex/`) and treat it as an extension of the static allowlist:

- `is_allowed_model` / `assert_allowed_model` accept catalog models with `supported_in_api`.
- `uses_responses_lite` honours the catalog's `use_responses_lite`, so a new Lite-lane model is not sent down the full Responses API.
- `-fast` aliases, `models --full`, the unknown-model hint and registry routing (`expand_codex_models`) include catalog models with `visibility == "list"`.

The static lists remain the baseline for machines without a cache. The file is parsed leniently per model entry and cached by `(mtime, len)`; an unreadable file keeps the last good parse.

## Verification

- `cargo test` (864 + integration suites) green, `cargo fmt --check` and `cargo clippy --all-targets -- -D warnings` clean.
- Manual: with a cache entry `gpt-7-test` (`supported_in_api: true`) the binary lists `gpt-7-test` and `gpt-7-test-fast` and routes requests to Codex without any code change; an entry with `supported_in_api: false` stays rejected.
- Live: `gpt-6-astra` requests through the patched binary against the Codex backend (Subscription/OAuth) succeed, including a 905k-token input.

### DIFF
--- a/src/providers/codex/chat_completions/request.rs
+++ b/src/providers/codex/chat_completions/request.rs
@@ -3,7 +3,7 @@ use serde_json::{Map, Value, json};
 
 use crate::providers::codex::translate::{
     model_allowlist::{
-        ALLOWED_MODELS, assert_allowed_model, resolve_model_request, uses_responses_lite,
+        allowed_models_display, assert_allowed_model, resolve_model_request, uses_responses_lite,
     },
     request::{Effort, resolve_effort_override, to_codex_effort},
 };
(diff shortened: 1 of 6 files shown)


Getty:
Good that this patch exists. Just bought OpenAI Pro subscription to use astra with claude code proxy and was a bit confused that i didnt got it! :D haha Thanks!

PR #133, opened 2026-09-07
feat(anthropic): add opt-in native passthrough

srnnkls:
Thanks for the feedback, and sorry my earlier description made the motivation hard to find.

## Why

I use the proxy for multiple backends, but sometimes need to switch a task back to native Claude. Today that means changing the client setup or bypassing the proxy, which also loses the proxy's single model-routing path and monitor history.

This adds an opt-in Anthropic provider. With `aliasProvider: anthropic`, Claude model IDs use the credentials already sent by Claude Code, while GPT, Kimi, Grok, OpenCode Go, and Cursor IDs keep their existing routes.

## What changes

- Relays the original Messages request and stream to Anthropic, preserving native fields and thinking blocks.
- Keeps requests visible in the existing monitor and handles token counting.
- Stores no Anthropic credentials; authentication stays with Claude Code.

Nothing changes for existing users unless they select `aliasProvider: anthropic`.

## Verification

Local mock-server tests cover buffered and streaming responses, errors, token counting, credential forwarding, and provider switching. The full Rust test suite and lint checks pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,12 @@ The opt-in Images API returns base64 image data and consumes the signed-in accou
 
 ## Providers
 
+To keep Claude models on Anthropic while switching GPT models through Codex in the
+same session, opt in with `CCP_ALIAS_PROVIDER=anthropic`. See
+[Anthropic passthrough](https://claude-code-proxy.raine.dev/providers/anthropic/)
+for subscription authentication and thinking compatibility.
+
+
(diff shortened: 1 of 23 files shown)


srnnkls:
Is anybody caring about PRs on this project?

raine:
If you want your PR merged, I'd recommend a different strategy. 

You could for example try a concise human readable explanation why you think it would be a good addition to the tool

PR #134, opened 2026-09-07
feat(monitor): attach dashboards to a background proxy

srnnkls:
## Summary

Running the proxy as a service currently loses the interactive monitor: the TUI owns the listener, while plain mode does not collect monitor history. This adds `claude-code-proxy monitor` to attach a dashboard to an existing proxy. Closing that dashboard leaves inference and other dashboards running.

- Keep monitor accounting in plain `serve --no-monitor` mode and handle Ctrl-C/SIGTERM with graceful service shutdown.
- Expose versioned, read-only snapshots at `GET /monitor`, restricted to actual loopback peers. Immutable read models carry elapsed durations and computed rates without reconstructing process-local clocks.
- Poll outside the terminal event loop, retain the last snapshot during outages, and reconnect automatically. `q` and Ctrl-C detach; dropping a dashboard cancels its polling task.

```sh
# Run under a service manager or in another terminal.
claude-code-proxy serve --no-monitor

# Attach one or more dashboards.
claude-code-proxy monitor
claude-code-proxy monitor --url http://127.0.0.1:19999
```

The existing integrated `serve` dashboard keeps its shutdown behavior. History remains in memory and resets on service restart. Remote hosts can be reached through an SSH port forward. The attached dashboard needs no provider credentials and has no server mutation or shutdown endpoint.

This branch is based directly on upstream `main` and is independent of #133.

## Validation

- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, and `cargo build --all-targets`.
- `cargo test --all-targets`, including snapshot serialization, loopback access checks, multiple viewers, detach/reattach, polling failure/recovery, cancellation, and incompatible protocol handling.
- Real pseudo-terminal smoke test: start the headless proxy, attach and quit with `q`, reattach and quit with Ctrl-C, verify HTTP health after each detach, then verify clean SIGTERM shutdown.

Also ran the edition and Clippy fixers. Validation uses local mock servers; it makes no live provider requests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,10 @@ Start the proxy in one terminal:
 claude-code-proxy serve
 ```
 
+For a background service, use `claude-code-proxy serve --no-monitor` and attach
+from another terminal with `claude-code-proxy monitor`. Closing an attached
+dashboard leaves the proxy running.
+
 Start Claude Code in another:
 
 ```sh
--- a/docs/src/content/docs/reference/command-reference.md
+++ b/docs/src/content/docs/reference/command-reference.md
@@ -30,6 +30,16 @@ Starts the local HTTP proxy and blocks until shutdown.
 
 The bind address comes from `CCP_BIND_ADDRESS` or `bindAddress`. Interactive stdout opens the monitor unless `--no-monitor` is present. Non-terminal stdout uses plain mode.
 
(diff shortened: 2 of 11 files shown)


raine:
Thanks for this! Being able to attach a dashboard to a background proxy is useful. Merged.

PR #138, opened 2026-09-08
fix(opencode): send x-opencode-session header to OpenCode Go upstream

### LINKED ISSUE #137: OpenCode Go rejects requests over missing `x-opencode-session`
Hi!

I've been using the proxy with OpenCode Go for a while now, and today I noticed I no longer can. Any attempt yields the following error

```
API Error: 400 Error from provider (Console Go): Request is missing x-opencode-session and cannot be routed efficiently. Please see https://opencode.ai/docs/go/#where-can-i-use-it
```

The issue seems to stem from OpenCode Go (opencode.ai/zen/go) having announced on 2026-09-03 that API requests must carry an `x-opencode-session` header (one stable ID per conversation), and requests missing it may start erroring from 2026-09-06.

Similar issues seem to have been reported in most harnesses repos like https://download.plaud.ai/openclaw/openclaw/issues/137165, https://github.com/earendil-works/pi/issues/4847, https://github.com/lidge-jun/opencodex/issues/3344.

I wonder if something could be done at the proxy level to circumvent the issue?

PauloPhagula:
## Summary
- OpenCode Go now rejects requests missing `x-opencode-session` ("Request is missing `x-opencode-session` and cannot be routed efficiently"), so all OpenCode Go traffic through the proxy currently fails.
- `OpenCodeClient::post` now sets `x-opencode-session` on every upstream request, using the Claude Code session id already threaded through `RequestContext.session_id` (from `x-claude-code-session-id`), falling back to a generated UUID for callers with no session id.
- Traffic-capture metadata now reflects the `x-opencode-session` value sent upstream.

Fixes #137

## Test plan
- [x] `cargo test opencode` — 41 tests pass, including new header-capture and end-to-end forwarding assertions
- [x] `cargo clippy --all-targets` — clean
- [x] `cargo fmt --check` — clean

### DIFF
--- a/src/providers/opencode/client.rs
+++ b/src/providers/opencode/client.rs
@@ -78,6 +78,7 @@ impl OpenCodeClient {
         body: &T,
         stream: bool,
         traffic: Option<Arc<TrafficCapture>>,
+        session_id: Option<&str>,
     ) -> Result<OpenCodeResponse, OpenCodeError> {
         let Some(api_key) = self.api_key.as_deref().filter(|key| !key.is_empty()) else {
             return Err(OpenCodeError {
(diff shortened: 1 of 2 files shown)


raine:
Thanks

PR #143, opened 2026-09-09
Honor SSL_CERT_FILE/SSL_CERT_DIR and the platform trust store in HTTP clients

chillosu:
## Summary
HTTP clients only trust the bundled webpki roots, so any host behind a private CA or a TLS-inspecting proxy gets `error sending request for url (...)` on every OpenAI request, and `SSL_CERT_FILE` has no effect. The websocket client already combines the platform trust store with webpki roots; this makes the HTTP clients do the same.

## Change
One Cargo feature: `reqwest` gains `rustls-tls-native-roots`. reqwest then loads the platform trust store, or the standard `SSL_CERT_FILE` / `SSL_CERT_DIR` locations when set, in addition to webpki roots. `rustls-native-certs` is already a dependency, so the lockfile change is feature-only.

## Verification
- Before: `claude-code-proxy codex auth device` behind a TLS-inspecting proxy → `Device init network error: error sending request for url (https://auth.openai.com/api/accounts/deviceauth/usercode)`; a CONNECT-tunnel capture shows the handshake complete and the client send a TLS alert (untrusted chain).
- After (this branch): the same command returns `Visit: https://auth.openai.com/codex/device` with a device code, with the inspection CA provided via `SSL_CERT_FILE`.
- `curl` with the same CA file verifies the same endpoint, confirming the chain is valid on the host.

## Notes
- Behaviour on hosts without a private CA is unchanged (webpki roots remain trusted).
- Happy to add a docs line under configuration: "Trust store: the platform store is used; set `SSL_CERT_FILE` or `SSL_CERT_DIR` to provide a custom CA bundle."


raine:
Thanks for adding this! HTTP clients now use the same native-root trust approach as WebSocket. Merged.

PR #145, opened 2026-09-10
fix(opencode): refresh Go catalog and accept terminal pings

Aser-Osama:
OpenCode Go's live catalog includes models that the proxy does not yet recognize, so Claude Code rejects those model IDs before reaching the provider. This refreshes the catalog with the 14 missing models and their Chat Completions, Messages, or Responses routes. It also keeps `grok-4.6` provider-qualified so the bare ID remains owned by the native Grok provider, and extends GLM reasoning-effort handling to GLM 5.3.

Live testing exposed a second issue on the newly routed `opencode-go/grok-4.6`: OpenCode appends a billing `ping` after `response.completed`, which made both buffered and streaming translation fail with HTTP 502. The translator now ignores only that harmless trailing ping while continuing to reject duplicate terminal events, late content, and malformed trailing frames.

Added models:

- `grok-4.6`, `glm-5.3`, `glm-5.3-flash`, `longcat-2.0`
- `deepseek-flash`, `deepseek-v4-flash-vision-exp`
- `mimo-v2-pro`, `mimo-v2-omni`
- `qwen3.8-flash`, `hy4-preview`, `hy3-preview`
- `muse-spark-1.3-contributor`, `muse-spark-1.2-contributor`, `omen-alpha`

Validation:

- `cargo test` — all unit, integration, and smoke suites pass
- Live GLM 5.3 Flash request through the proxy — HTTP 200
- Live Qwen 3.8 Flash request through the proxy — HTTP 200
- Live Grok 4.6 buffered and streaming requests through the proxy — HTTP 200 with clean completion

### DIFF
--- a/docs/src/content/docs/providers/opencode-go.md
+++ b/docs/src/content/docs/providers/opencode-go.md
@@ -34,9 +34,9 @@ ANTHROPIC_SMALL_FAST_MODEL=opencode-go/glm-5.2 \
   claude --model opencode-go/glm-5.2
 ```
 
-The bare IDs `gpt-5.6-luna`, `grok-4.5`, `kimi-k3`, and `kimi-k2.6` remain
-owned by the existing Codex, Grok, or Kimi providers. Prefix those IDs with
-`opencode-go/` to select the OpenCode Go version.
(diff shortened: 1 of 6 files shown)


raine:
Thank you

PR #150, opened 2026-09-13
isolate codex transport tests from the developer's config

rikbrown:
## Problem

The five `config::tests::codex_transport_*` tests call `clear_env()`, which removes `CCP_CONFIG_DIR`, and then call `codex_transport()` without pointing it at a temp dir. The accessor falls back to `paths::config_dir()`, so the tests read the developer's real `~/.config/claude-code-proxy/config.json`.

On any machine whose config sets `codex.transport` to something other than the default, `cargo test` fails:

```
test config::tests::codex_transport_defaults_to_websocket ... FAILED
test config::tests::codex_transport_empty_env_falls_back_to_websocket ... FAILED
test config::tests::codex_transport_env_websocket ... FAILED
test config::tests::codex_transport_invalid_env_falls_back_to_websocket ... FAILED
test config::tests::codex_transport_reads_env ... FAILED
```

Only one of those is a real assertion failure — the shared `ENV_LOCK` is then poisoned and takes the other four down with it, so a single config value presents as five failures.

Worth noting: `CCP_CONFIG_DIR=$(mktemp -d) cargo test` does **not** work around this, because `clear_env()` strips exactly that variable. The suite has to be isolated via `HOME`, which is not obvious.

## Reproduce

```sh
mkdir -p /tmp/ccp-repro/.config/claude-code-proxy
printf '{"codex":{"transport":"auto"}}' > /tmp/ccp-repro/.config/claude-code-proxy/config.json
HOME=/tmp/ccp-repro XDG_CONFIG_HOME=/tmp/ccp-repro/.config cargo test config::tests::codex_transport
```

## Fix

Point each of the five tests at a `tempfile::TempDir` via `EnvGuard::set("CCP_CONFIG_DIR", ...)` — the same shape `codex_server_compaction_defaults_and_overrides` and the other config accessor tests in the module already use. Assertions are unchanged, and `clear_env()` is deliberately left alone since other tests rely on the unset state.

I checked every other `clear_env()` caller in the module; they all either set `CCP_CONFIG_DIR` already or use `load_config_for_env` with an explicit dir, so this is the only instance.

## Verification

- Repro above: 5 failed before, `6 passed; 0 failed` after
- `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt --check` clean
- Full suite with an isolated `HOME`: green


raine:
Thanks for catching this! Merged.

PR #151, opened 2026-09-13
apply the compaction effort cap to requests naming no effort

rikbrown:
## Problem

`src/providers/codex/translate/request.rs`:

```rust
if apply_codex_config
    && is_compact
    && let Some(cap) = compact_effort_cap()
    && resolved_effort.as_ref().is_some_and(|e| *e > cap)
{
    resolved_effort = Some(cap);
}
```

`is_some_and` means the cap only ever *lowers* an effort that is already present. A compaction request naming no effort at all skips the cap entirely and runs at the upstream default — which is the effort level the cap exists to avoid.

That is also the case it is most needed for: Claude Code's own `/compact` request carries no effort of its own, so the fast path never applied to it.

## Fix

`apply_compact_effort_cap` resolves the two inputs together:

| resolved effort | cap | result |
|---|---|---|
| none | `Low` | `Low` |
| `None` | `Low` | `None` (explicit, below cap) |
| `Low` | `Low` | `Low` |
| `High` | `Low` | `Low` (lowered) |
| anything | cap off | unchanged |

The existing invariant holds: the cap never raises effort, and an explicitly lower effort is preserved. The doc comment on `compact_effort_cap` is updated to state that a missing effort now takes the cap too.

## Verification

- New test `compact_effort_cap_defaults_a_missing_effort` covers every row above
- Existing `compact_request_downgrades_effort_to_cap` and `compact_effort_cap_parses_env_values` unchanged and passing
- `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt --check` clean, full suite green


raine:
Thanks for catching the missing-effort case! Merged.

PR #152, opened 2026-09-13
raise the Anthropic request body limit to 64 MiB

rikbrown:
## Problem

`MAX_OPENAI_REQUEST_BYTES` is 16 MiB. An image-heavy Claude Code session (~490k tokens of screenshots) pushes the request body past that, and because Claude Code re-sends the full history on every turn, the gate then rejects **every subsequent request on that session** — the session is unrecoverable without dropping the transcript.

## Fix

Raise the limit to 64 MiB, matching the Codex image edit cap already used in `providers::codex::images`, so the two paths agree on how large an image-bearing request may be.

`MAX_PROVIDER_STREAM_BYTES` (32 MiB) and `MAX_SSE_EVENT_BYTES` (2 MiB) are untouched — those bound responses, which this does not change.

## Verification

Tests in `tests/server.rs` cover the boundary directly:

- a body at the limit clears the gate
- a body over the limit still returns the length error
- a body over the old 16 MiB limit now clears the gate

`cargo clippy --all-targets -- -D warnings` clean, `cargo fmt --check` clean, full suite green.


RichAyotte:
Raising `MAX_OPENAI_REQUEST_BYTES` also raises `/v1/chat/completions` and `/v1/responses`, which still read that constant.

`messages_body_over_limit_returns_length_error` still expects 400 with `length limit exceeded`. Anthropic's Messages and Token Counting APIs document 413 `request_too_large` for overflow (https://platform.claude.com/docs/en/api/errors#request-size-limits).

raine:
Thanks for this! The larger limit fixes the image-heavy session case nicely. I split the Anthropic limit from the OpenAI-compatible endpoints and corrected oversized Anthropic requests to return 413 request_too_large. Merged.

PR #153, opened 2026-09-13
redact encrypted_content in traffic captures

rikbrown:
## Problem

Codex server-side compaction returns an opaque `encrypted_content` blob, which the proxy replays as a `compaction` input item on every later turn. With `CCP_TRAFFIC_LOG=1` that blob is written to disk verbatim.

To be clear about severity: the blob is ciphertext we cannot read locally, so this is **not** a plaintext leak. But it is a large, replayable handle to a whole conversation's context, and because it is re-sent every turn it accumulates across a session's captures.

`image_url` and Anthropic `source.data` already have explicit arms in `redact_traffic` for the same reason — bulk conversation payload does not belong in a debug capture. This applies the existing precedent to `encrypted_content`.

## Fix

A new arm in `redact_traffic_with_depth`, next to the `source.data` case, routing the value through the existing `redact_traffic_value` so it becomes `[redacted len=N]`. Blob size stays visible, since that is genuinely useful diagnostic information.

**Not** added to `REDACT_KEYS`, for two reasons:

1. `REDACT_KEYS` is shared with the structured logger via `logging::redact_value`, which is only ever called on the `fields` map in `Logger::emit`. No body, input item, or event value reaches the logger, so adding the key there would be a no-op for logging.
2. `REDACT_KEYS` is a credential list. This is bulk payload, which belongs with the `image_url` / `source.data` arms.

Reasoning items carry the same key and are covered by the same arm.

## Scope — what this does not cover

This covers `020-upstream-request` (`write_json`) and the `040`/`050` per-event JSON captures (`write_json_event`).

It does **not** cover the raw `032-upstream-response-body.sse` capture on the Codex HTTP and WebSocket paths (`client.rs`, `websocket.rs`), which is written with `write_bytes` and bypasses redaction entirely. The blob still persists there. Fixing that means either re-serialising parsed events or a text-level scrub of the SSE, which felt like a separate change — happy to follow up if you'd prefer it in one go.

Providers that assemble `upstream_sse` from `StreamTrafficCapture::upstream_event` (grok, opencode) are already redacted, since those frames go through `write_json_event`. Codex is the one path that dumps raw bytes.

## Verification

- `redact_traffic_strips_compaction_encrypted_content` — a `compaction` item in a request body; asserts the blob is replaced by `[redacted len=N]`, the `gAAAAA` prefix is gone, and `type`, message text and `model` survive
- `redact_traffic_strips_reasoning_encrypted_content_in_events` — same for a reasoning item inside `response.output_item.done`
- `cargo clippy --all-targets -- -D warnings` clean, `cargo fmt --check` clean, full suite green


raine:
Thanks for the redaction fix! Merged, including the reasoning-signature capture case.

PR #154, opened 2026-09-13
fix(grok): seed stream usage so Claude Code's bar is not zero

RichAyotte:
Grok streams left `message_start.usage.input_tokens` at 0 and omitted `input_tokens` from the terminal `message_delta`. Claude Code copies that delta field only when it is present, so the status bar stayed at zero input tokens after every Grok turn.

The stream now seeds `message_start` from the local token estimate and writes the provider's `input_tokens` on the finish delta, matching the Codex path. Docs also record that Grok 4.5/4.6 are 500k tokens and that Claude Code needs `CLAUDE_CODE_MAX_CONTEXT_TOKENS=500000` for those IDs.


raine:
Thanks for fixing the Grok usage display! Merged.

PR #156, opened 2026-09-17
fix(codex): use session-id header for cache affinity

Getty:
## Summary

- send the Codex Responses cache-affinity header as `session-id` instead of `session_id`
- keep `prompt_cache_key`, `x-client-request-id`, and `x-codex-window-id` behavior unchanged
- update the existing header tests and explicitly verify that the underscore variant is no longer emitted

## Why

`build_codex_headers` currently emits `session_id`. The official OpenAI Codex client emits `session-id`, and the current Codex source notes that ChatGPT derives Responses cache affinity from that header.

The proxy already supplies a stable `prompt_cache_key` from the Claude Code session ID. This patch therefore fixes the missing cache-routing signal without changing request-body semantics, continuation state, or session ownership.

Related upstream investigation: `openai/codex#44716` reports that `session-id` can affect reported cache reuse independently of `prompt_cache_key` on the ChatGPT-authenticated Codex backend. The backend mechanism is still under discussion there, so this PR intentionally stays limited to matching the official Codex wire format.

## Scope

This does **not** add `thread-id`, `prompt_cache_options`, or cache breakpoints. It also does not change the existing per-agent continuation ownership logic. Those are separate behavior changes and should be evaluated independently.

## Testing

- `cargo fmt --check`
- `cargo test codex_headers_`


raine:
Thanks for catching this and for the careful upstream investigation. I updated the continuation integration coverage for `session-id` and removed the stronger cache implementation claim. The current Codex source and header history confirm the hyphenated form is the right wire format.

Merged, thanks!

